PR #1, opened 2017-02-02
Update build instructions

csjperon:
- Update build instructions so people know where the .mk files are


csjperon:
Add a note that the --recursive clone flag is required for the .mk files

katef:
Thanks!

PR #2, opened 2017-02-02
bugfix and minor optimization for subsetof

marchelzo:
save some time by ignoring the initial segment of `b` where we know the next element of `a` cannot be.

also make it so that `subsetof(NULL, b)` is true for all `b`.


katef:
Thank you!

PR #6, opened 2017-02-22
lx, libre: Add support for inputs with embedded 0 bytes

dhobsd:
Previously, lx would generate incorrect output when literals or regexps contained `0` bytes as it used `xstrdup`, which would truncate at the byte. This caused other failures when the first byte was a `\0`.

After this, libre would also fail to handle such regular expressions during compile due to lx consuming its getc interface that expects C strings.

### DIFF
--- a/src/libre/getc.c
+++ b/src/libre/getc.c
@@ -39,3 +39,18 @@ re_fgetc(void *opaque)
 	return fgetc(f);
 }
 
+int
+re_bgetc(void *opaque)
+{
+	struct re_buf *b = opaque;
+
+	assert(b != NULL);
+	assert(b->ptr != NULL);
+	assert(b->pos <= b->len);
+
+	if (b->pos == b->len) {
+		return EOF;
+	}
+
+	return (unsigned char) b->ptr[b->pos++];
+}
--- a/src/lx/parser.act
+++ b/src/lx/parser.act
@@ -45,7 +45,12 @@
 	#include "out.h"
 	#include "var.h"
 
-	typedef char *               string;
+	struct string {
+		char *p;
+		size_t l;
+	};
+
+	typedef struct string *      string;
 	typedef enum re_flags        flags;
 	typedef struct fsm *         fsm;
 	typedef struct ast_zone *    zone;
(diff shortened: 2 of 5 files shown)


src/libre/getc.c
```diff
@@ -39,3 +39,16 @@ re_fgetc(void *opaque)
 	return fgetc(f);
 }
 
+int
+re_bgetc(void *opaque)
+{
+	struct re_buf *b = opaque;
+
+	assert(b != NULL);
```
katef: Better assert that `b->ptr != NULL` here

katef: and that `b->pos <= b->len`

src/lx/parser.act
```diff
@@ -194,28 +192,37 @@
...
 			exit(EXIT_FAILURE);
 		}
+		@s->l = strlen(@s->p);
 	@};
 
 	IDENT: () -> (s :string) = @{
-		@s = xstrdup(lex_state->buf.a);
+		@s = malloc(sizeof *@s);
 		if (@s == NULL) {
 			perror("xstrdup");
```
katef: "malloc"

src/lx/parser.act
```diff
@@ -245,12 +252,20 @@
 		 * and so the pattern buffer will be overwritten by the LL(1) one-token
 		 * lookahead.
 		 */
```
katef: This comment seems out of date

src/lx/parser.act
```diff
@@ -350,17 +370,20 @@
 
 	<compile-regex>: (s :string, f :flags) -> (r :fsm) = @{
 		struct re_err err;
-		const char *s;
+		struct re_buf buf;
```
katef: Would it work out neater to use `struct re_but` for the `:string` type? I'm not sure.

dhobsd: Thanks for this comment. I was thinking about it, but we don't need the offset bit for this and these live in automatic storage anyway. Also, string is used by things that aren't RE_NATIVE/RE_LITERAL regexps, and so that felt a little gross. I think in general the allocation overhead for the `struct string` will dwarf the copy overhead into automatic storage here.

So that's why I didn't do it.

katef: Fine by me

PR #9, opened 2017-02-25
set: comparison callbacks and improvements to search/iteration

dhobsd:
* Add a comparison function to set. This can be set by calling a new
   function, `set_create`, which builds the set and assigns a comparison
   function.

 * Make `set_contains` a useful search function, able to return the item
   it found (if it found one).

 * Add `set_firstafter`, an iterator function that returns the next
   element logically after the element provided.

### DIFF
--- a/src/adt/set.c
+++ b/src/adt/set.c
@@ -13,6 +13,7 @@ struct set {
 	void **a;
 	size_t i;
 	size_t n;
+	int (*cmp)(const void *, const void *);
 };
 
 static void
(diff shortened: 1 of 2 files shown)


src/adt/set.c
```diff
@@ -69,6 +68,35 @@ set_search(const struct set *set, const void *item)
...
+{
+
+	return (a > b) - (a < b);
+}
+
+struct set *
+set_create(int (*cmp)(const void *a, const void *b))
+{
+	struct set *s;
+
```
katef: `assert(cmp != NULL);` here please, and also in whatever functions use it

PR #11, opened 2017-02-25
libre, lx: move bgetc implementation into lx's parser

dhobsd:


include/re/re.h
```diff
@@ -155,7 +149,7 @@ re_perror(enum re_dialect dialect, const struct re_err *err,
  */
 int re_sgetc(void *opaque); /* expects opaque to be char ** */
 int re_fgetc(void *oapque); /* expects opaque to be FILE *  */
-int re_bgetc(void *opaque); /* expects opaque to be struct re_buf * */
+int re_bgetc(void *opaque); /* expects opaque to be struct string * from adt/string.h */
```
katef: This can't be part of the public API unless `struct string` is, too.

dhobsd: But it is? For `re_fgetc`, you have to `#include <stdio.h>`, for `re_bgetc` you have to `#include <adt/string.h>`.

dhobsd: Oh, I see, we don't install `include/adt`'s headers.

PR #14, opened 2017-02-27
libfsm: compact storage for edges

dhobsd:
This stores edges using adt/set. It should result in less memory usage
in most cases, as well as faster runtime in cases like minimization,
which iterate over all edges over all states: now such cases can ignore
empty edges entirely.

### DIFF
--- a/src/adt/set.c
+++ b/src/adt/set.c
@@ -338,18 +338,16 @@ set_firstafter(const struct set *set, struct set_iter *it, void *item)
 
 	i = set_search(set, item);
 	r = set->cmp(item, set->a[i]);
-	if (i == 0) {
-		if (r < 0) {
-			it->i = 0;
-		} else {
-			it->i = 1;
-		}
-	} else {
-		if (r < 0) {
-			it->i = i;
-		} else {
-			it->i = i + 1;
-		}
+	assert(i <= set->i - 1);
+
+	if (r >= 0 && i == set->i - 1) {
+		it->set = NULL;
+		return NULL;
+	}
+
+	it->i = i;
+	if (r >= 0) {
+		it->i++;
 	}
 
 	it->set = set;
--- a/src/libfsm/internal.h
+++ b/src/libfsm/internal.h
@@ -7,6 +7,7 @@
 
 #include <fsm/fsm.h>
 
+struct set;
 
 /* TODO: +2 for SOL, EOL */
 /* TODO: +lots for FSM_EDGE_* */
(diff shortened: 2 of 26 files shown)


src/adt/set.c
```diff
@@ -74,7 +74,7 @@ static int
 set_cmp(const void *a, const void *b)
 {
 
-	return (a > b) - (a < b);
+	return (a < b) - (a > b);
```
katef: Hah!

katef: Surely this was right as it was... don't we want to return +1 when `a > b` and -1 when `a < b`?

src/libfsm/internal.h
```diff
@@ -18,12 +19,13 @@ enum fsm_edge_type {
...
 struct fsm_edge {
 	struct set *sl;
+	enum fsm_edge_type symbol;
 };
 
 struct fsm_state {
 	unsigned int end:1;
 
-	struct fsm_edge edges[FSM_EDGE_MAX + 1];
+	struct set *edges;
```
katef: Better comment about what type the set's `void *` points to.

src/libfsm/minimise.c
```diff
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <stddef.h>
 #include <stdlib.h>
+#include <stdio.h>
```
katef: Left over from debugging?

src/libfsm/mode.c
```diff
@@ -26,18 +26,31 @@ fsm_findmode(const struct fsm_state *state, unsigned int *freq)
...
 
 			curr = 0;
 
-			/* count the remaining edes which have the same target */
-			for (j = i + 1; j <= UCHAR_MAX; j++) {
-				curr += set_contains(state->edges[j].sl, s);
+			/* Count the remaining edes which have the same target.
+			 * This works because the edges are still sorted by
+			 * symbol, so we don't have to walk the whole thing.
+			 */
```
katef: Glad you commented about that!

src/libfsm/state.c
```diff
@@ -9,6 +9,20 @@
...
+{
+	const struct fsm_edge *ea, *eb;
+
+	assert(a != NULL);
+	assert(b != NULL);
+
+	ea = a;
+	eb = b;
+
+	return (ea->symbol > eb->symbol) - (ea->symbol < eb->symbol);
```
katef: If the ordering matters, better comment about it (I think we rely on it, during iteration)

src/adt/set.c
dhobsd: We didn't regenerate lexers after this change, so we didn't see. But it changed the ordering of states. I don't think it's a big deal, but it was making it annoying for me to test on a machine that didn't have sid installed.

dhobsd:
Please hold on merging, libfsm cannot build itself from this version.

PR #16, opened 2017-02-28
libfsm: turn transset and ml into sets

dhobsd:


src/libfsm/determinise.c
```diff
@@ -40,69 +40,84 @@ struct mapping {
 
 	/* boolean: are the outgoing edges for dfastate all created? */
 	unsigned int done:1;
-
-	struct mapping *next;
 };
 
+static struct set *trans = NULL;
+static struct set *mappings = NULL;
```
katef: Oh no... I want to keep libfsm thread safe.

PR #20, opened 2017-03-02
libfsm: fix carryopaque in fsm_reverse

dhobsd:


katef:
I committed something else which does the same as this fix; thanks!

PR #29, opened 2017-03-24
`fsm_reverse()` in place

edk0:
I don't think the code is nice enough to merge yet.

This slightly changes some API: if `fsm_reverse(x)` fails, `x` is no longer guaranteed to be unchanged, but still needs to be freed.

### DIFF
--- a/src/libfsm/reverse.c
+++ b/src/libfsm/reverse.c
@@ -63,11 +63,16 @@ fsm_reverse(struct fsm *fsm)
 				struct fsm_state *to;
 				struct set_iter jt;
 				for (to = set_first(e->sl, &jt); to != NULL; to = set_next(&jt)) {
-					/*XXX */
 					if (lq <= nq) {
+						void *tmp;
 						lq = lq > 0 ? lq * 2 : 1024;
-						q = realloc(q, lq * sizeof *q);
-						if (!q) abort();
+						tmp = realloc(q, lq * sizeof *q);
+						if (!tmp) {
+							free(q);
+							set_free(endset);
+							return 0;
+						}
+						q = tmp;
 					}
 					/* reversed because it's a description of the edge we wish to create later */
 					q[nq].from = to;
--- a/src/libfsm/reverse.c
+++ b/src/libfsm/reverse.c
@@ -6,6 +6,7 @@
 
 #include <assert.h>
 #include <stddef.h>
+#include <stdlib.h>
 
 #include <adt/set.h>
 
@@ -20,23 +21,17 @@
 int
 fsm_reverse(struct fsm *fsm)
 {
-	struct fsm *new;
 	struct fsm_state *end;
 	struct set *endset;
 
 	assert(fsm != NULL);
 	assert(fsm->opt != NULL);
 
-	new = fsm_new(fsm->opt);
-	if (new == NULL) {
-		return 0;
-	}
-
 	/*
 	 * The new end state is the previous start state. Because there is (at most)
 	 * one start state, the new FSM will have at most one end state.
 	 */
-	end = NULL;
+	end = fsm->start;
 
 	/*
 	 * The set of the previous end states.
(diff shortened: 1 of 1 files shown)


src/libfsm/reverse.c
```diff
@@ -47,86 +42,74 @@ fsm_reverse(struct fsm *fsm)
...
-			if (s == fsm->start) {
-				end = p;
-				fsm_setend(new, end, 1);
+		for (s = fsm->sl; s != NULL; s = s->next) {
+			struct fsm_edge *e;
+			for (e = set_first(s->edges, &it); e != NULL; e = set_next(&it)) {
+				struct fsm_state *to;
+				struct set_iter jt;
+				for (to = set_first(e->sl, &jt); to != NULL; to = set_next(&jt)) {
+					/*XXX */
```
katef: Why the XXX?

src/libfsm/reverse.c
```diff
@@ -47,86 +42,74 @@ fsm_reverse(struct fsm *fsm)
...
+						void *tmp;
+						lq = lq > 0 ? lq * 2 : 1024;
+						tmp = realloc(q, lq * sizeof *q);
+						if (!tmp) {
+							free(q);
+							set_free(endset);
+							return 0;
+						}
+					}
+					/* reversed because it's a description of the edge we wish to create later */
```
katef: Let's store `q[]` with `.from` and `.to` in the current direction, and pass those the other way around to fsm_addedge() instead.

src/libfsm/reverse.c
edk0: Oops, left over from when that could abort()

src/libfsm/reverse.c
```diff
@@ -47,86 +42,72 @@ fsm_reverse(struct fsm *fsm)
...
-	{
-		struct fsm_state *s;
 
-		for (s = fsm->sl; s != NULL; s = s->next) {
-			struct fsm_state *to;
-			struct fsm_state *se;
-			struct set_iter it;
-			struct fsm_edge *e;
-
-			to = s->equiv;
```
katef: Nice that we don't need `.equiv` here any more!

src/libfsm/reverse.c
```diff
@@ -47,86 +42,72 @@ fsm_reverse(struct fsm *fsm)
...
+				for (to = set_first(e->sl, &jt); to != NULL; to = set_next(&jt)) {
+					if (lq <= nq) {
+						void *tmp;
+						lq = lq > 0 ? lq * 2 : 1024;
+						tmp = realloc(q, lq * sizeof *q);
+						if (!tmp) {
+							free(q);
+							set_free(endset);
+							return 0;
+						}
```
katef: (I added this when testing)

src/libfsm/reverse.c
```diff
@@ -47,86 +42,72 @@ fsm_reverse(struct fsm *fsm)
...
-
-			for (e = set_first(s->edges, &it); e != NULL; e = set_next(&it)) {
-				struct set_iter jt;
-
-				for (se = set_first(e->sl, &jt); se != NULL; se = set_next(&jt)) {
-					struct fsm_state *from;
-					struct fsm_edge *edge;
-
-					assert(se != NULL);
+			/* deliberately don't go via the API. we'll correct the end count when we're finished with it. */
```
edk0: I don't want to change fsm->endcount. Probably better to save the original count instead.

src/libfsm/reverse.c
edk0: Oops!

src/libfsm/reverse.c
```diff
@@ -242,8 +214,7 @@ fsm_reverse(struct fsm *fsm)
 
 	set_free(endset);
 
-	fsm_move(fsm, new);
+	fsm->endcount = end != NULL ? 1 : 0;
```
katef: This can go now too, can't it?

src/libfsm/reverse.c
```diff
@@ -47,86 +44,74 @@ fsm_reverse(struct fsm *fsm)
...
-					struct fsm_edge *edge;
-
-					assert(se != NULL);
+			fsm_setend(fsm, s, s == end);
+		}
 
-					from = se->equiv;
+		/* the FSM now has no edges; add a reversed edge for each one recorded */
+		for (i = 0; i < nq; i++) {
+			fsm_addedge(q[i].to, q[i].from, q[i].symbol);
```
katef: Don't forget the error handling here

katef:
I think this is a little misplaced after all. Here we're removing all the edges in an FSM, and then re-creating them in the same FSM. That avoids re-creating the FSM itself, but that's not particularly expensive.

Instead, I'd like to try re-using the same edges, to avoid the overhead for constructing each edge. This should be a lot more straightforward when we have the internal representation converted to an adjacency list, so let's try this idea when that's present.

PR #32, opened 2017-07-10
Installs adt/set.h so external things can use carryopaque

sfstewman:


katef:
I don't want to have `include/set` installed, because that's a pretty obnoxious namespace to be populating. I'm not sure quite what I'd prefer, though. Thoughts welcome.

katef:
Perhaps for carryopaque's sake, this could be exposed as an array of however many `struct fsm_state`? Although I'd want to check @dhobsd is happy with the performance hit for constructing it. But at least in the case of a single item, we could just point at that one `struct fsm_state` without allocating anything to hold an array.

dhobsd:
Can we just change the namespace, then? I needed `adt/set.h` for the IP range thing, and I really don't have any great ideas for how to abstract it out. I also don't have any great ideas for improving the performance here because the real problem is the algorithmic complexity. Still, if we can do less work, that's better. `set` has a robust API now, and if we want to avoid polluting some `set_` namespace, it seems like it is maybe just a naming problem?

katef:
I think I'd rather have it present an array, which is what I was trying to describe above. Then we don't need to expose any of the set.h functions. That works out because `carryopaque()` only needs a read-only view.

katef:
Implemented by #38

PR #35, opened 2017-07-18
Add SQL99 regexp dialect

katef:
This is an implementation of the `SIMILAR TO` regexp dialect per ISO/IEC 9075-2:1999 (E)

More legible overview: https://firebirdsql.org/refdocs/langrefupd25-similar-to.html

Notable aspects:
* No `^` `$` predicate anchors
* Very few constructs; no `?` or `{m,n}` operators
* Both `ANY` and `MANY`, spelled `_` (as opposed to the more usual `.`) and `%`. The former corresponds to SQL's `LIKE` patterns; the latter does not typically exist in regexp dialects
* Character classes

Not implemented:
* Arbitrary user-supplied optional escape character
* Collation between character sets


man/sql.5e/sql.5e.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
edk0: Filename looks like a typo of `5re`

PR #36, opened 2017-07-18
fix typo

Erkajp:


dhobsd:
Thank you for submitting a patch! This project uses British English spellings in documentation and source code, so this is not a typo. I think you'd likely find many more similar instances throughout the source and documentation. As such, I think this is a "wontfix" kind of item.

Erkajp:
i'm sorry, thank you for nice reply

PR #39, opened 2017-07-21
Bugfix: Cast for `char` range

katef:
Character literals are escaped with the form `'\xff'`. These are converted to an integer value as if by a cast through `char` (the signedness of which is implementation-dependent).

In libfsm's generated code, we read a positive character value from lx's getc interface (as if cast to `unsigned char`, much as fgetc has); when switching on that value, these positive then compare inequal to the character literals above where `char` is signed.

This PR casts to `char` when switching, to give both sides the same range. An alternative fix would be to emit integer literals of the form `0xff` instead of character literals, but I prefer this cast because then a C compiler can potentially reason about the range being switched on, which is a char rather than an int.


sfstewman:
> This PR casts to char when switching, to give both sides the same range.

For signed char, isn't this behavior IDB for values outside of its range, which allows an implementation to do anything (such as floor the out-of-range values to zero) or raise a signal?

> These are converted to an integer value as if by a cast through char (the signedness of which is implementation-dependent).

Why aren't they converted to an int as if cast through an unsigned char, the way the C stdlib tends to handle this?

katef:
The "These are converted to an integer value as if by a cast through char" part refers to C's behaviour for `\xff` inside character literals. That's not something we can control.

edk0:
I believe that "cast through char" is not an entirely accurate explanation. The standard seems to say that character constants start with a value in `char` corresponding to the content of the character constant. So I think that `'\xff'` is well-defined, even on signed char platforms, and it's only the conversion in the switch expression we need to worry about; a simple manual two's-complement would do if we're worried about portability.

PR #43, opened 2017-07-24
Adds a void * argument to predicates

sfstewman:


include/fsm/fsm.h
```diff
@@ -156,8 +156,8 @@ fsm_getopaque(struct fsm *fsm, const struct fsm_state *state);
  * Returns NULL on error, or if there is no state.
  */
 struct fsm_state *
-fsm_collate(struct fsm *fsm,
-	int (*predicate)(const struct fsm *, const struct fsm_state *));
+fsm_collate(struct fsm *fsm, void *pred_arg,
+	int (*predicate)(void *, const struct fsm *, const struct fsm_state *));
```
katef: Name it `void *opaque` please, and (sorry to be annoying), move it to the last argument for `(*predicate)()`

katef:
After a bit more thought, here's why I'm uneasy about this:
1. Predicates are only intended to be executed when necessary; if some node-walking function is able to show it doesn't need to visit a certain sub-graph, then there's no guarantee a predicate would be called for any node in that sub-graph.
2. From (1), the purpose of a predicate is to supply information to the node-walking functions. The only user-supplied information currently attached to an FSM is (purposefully) at nodes marked as end states. Those have a `void *` for that reason, which user-supplied predicates may inspect.
3. Your use-case is iteration, not search sub-graphs.

(1) and (2) are why I'm uneasy about adding this feature for predicates. And I think (3) is enough to say that these predicate callbacks aren't the right tool for what you want.

So instead, I suggest something quite similar to what you'd originally said: to provide an "iterate this callback over every node" function, passed a callback which does take a void *, and isn't named "predicate" (I'd just call it `f`). We wouldn't supply any pre-made callbacks. There could be an interface for iterating over edges, too.

Now those iteration functions would _also_ take an optional predicate pointer, and there's not necessarily a user-visible relation between when that's called vs. when the user's `f` is called.

PR #45, opened 2017-07-25
Initial PCRE support

edk0:
Add the `pcre` dialect, featuring the following features:

- Properly scoped `(?i)`/`(?-i)`
- Empty alts `(hats|)`
- Non-capturing groups (superficially: we don't have capturing yet)


src/libre/parser.act
```diff
@@ -988,6 +998,29 @@
 	@};
 
 
+	<push-flags>: () -> () = @{
+		struct flags *n = malloc(sizeof *n);
```
katef: deal with NULL here

PR #49, opened 2017-08-01
Transition table

sfstewman:
Adds a programmatic interface to convert fsm to/from a table of transitions (adjacency list), and adds the ability to output the table in an ascii format and to read the table into the fsm tool.  Also adds some tests.


include/fsm/walk.h
```diff
@@ -49,5 +49,21 @@ int
...
+fsm_walk_states(const struct fsm *fsm, void *opaque,
+	int (*callback)(const struct fsm *, const struct fsm_state *, void *));
+
+/* Allows iterating through the states of the graph with a callback
+ * function.  Takes an opaque pointer that the callback can use for its
+ * own purposes.
+ */
+int
+fsm_walk_edges(const struct fsm *fsm, void *opaque,
+	int (*callback)(const struct fsm *, const struct fsm_state *, unsigned int, const struct fsm_state *, void *));
```
katef: Could I have just these in their own PR please?

katef:
Per discussion on IRC, we're considering moving the internal representation for FSM to use an adjacency list in the first place, instead.

PR #51, opened 2017-08-02
Adds functions to query number of states and edges

sfstewman:


src/libfsm/fsm.c
```diff
@@ -129,3 +129,46 @@ fsm_carryopaque(struct fsm *fsm, const struct set *set,
...
+	for (src = fsm->sl; src != NULL; src = src->next) {
+		struct set_iter ei;
+		const struct fsm_edge *e;
+
+		for (e = set_first(src->edges, &ei); e != NULL; e=set_next(&ei)) {
+			struct set_iter di;
+			const struct fsm_state *dst;
+			for (dst = set_first(e->sl, &di); dst != NULL; dst=set_next(&di)) {
+				assert(n+1>n); /* handle overflow with more grace? */
+				n++;
```
katef: There's a `set_count()` we could use here.

PR #52, opened 2017-08-04
Fix bug caused by unintended sign extension.

silentbicycle:
The callback passed to `fsm_exec` returns int, to allow for EOF, but
the code paths that handled the result were inconsistent with the
sign of that value after checking it for EOF. In particular, characters
>= 0x80 could be sign-extended to 0xffffff80 between
`state = nextstate(state, c)` and `nextstate`'s call to
`fsm_hasedge`, which would lead to state match failures in an otherwise
correctly constructed FSM.

(theft found this bug.)

Note: I haven't been able to run the tests, they seem to depend on sid,
and I'm not sure this hasn't broken something else. In particular,
it may break handling of epsilon, which is defined as the max unsigned
char + 1.

### DIFF
--- a/src/libfsm/exec.c
+++ b/src/libfsm/exec.c
@@ -18,7 +18,7 @@
 #include "internal.h"
 
 static struct fsm_state *
-nextstate(const struct fsm_state *state, char c)
+nextstate(const struct fsm_state *state, int c)
 {
 	struct fsm_edge *e;
 


silentbicycle:
theft says this isn't good enough: apparently it still fails when the first byte is 0xFF.

src/libfsm/exec.c
```diff
@@ -18,7 +18,7 @@
 #include "internal.h"
 
 static struct fsm_state *
-nextstate(const struct fsm_state *state, char c)
+nextstate(const struct fsm_state *state, unsigned char c)
```
katef: I'd rather `nextstate()` have its argument as an unqualified `char`, because that's currently what edges in FSMs are. and it's up to the caller to sort out the right if they're coming from some other type.

src/libfsm/exec.c
```diff
@@ -53,7 +53,7 @@ fsm_exec(const struct fsm *fsm,
 	assert(fsm->start != NULL);
 	state = fsm->start;
 
-	while (c = fsm_getc(opaque), c != EOF) {
+	while (c = fsm_getc(opaque), c != (unsigned int)EOF) {
```
katef: `c` itself should be declared `int c;` and no casts here. Just like when using the standard `getc()`.

src/libfsm/exec.c
silentbicycle: `enum fsm_edge_type` in `src/libfsm/internal.h` defines `FSM_EDGE_EPSILON` as `UCHAR_MAX + 1`, so that won't fit in a char either way.

The sign extension bug seems to be happening truncating from an int to a char, and then expanding to the enum. If `nextstate`'s `c` argument is instead an `enum fsm_edge_type`, then the conversion from an int is correct.

katef:
Thanks!

PR #53, opened 2017-08-07
Fix memory leaks: the sets in `fsm_determinise` were not freed.

silentbicycle:
This eliminates a consistent memory leak per call to `fsm_determinise`.


src/libfsm/determinise.c
```diff
@@ -607,6 +607,14 @@ fsm_determinise_freecache(struct fsm *fsm, struct fsm_determinise_cache *dcache)
 	free_mappings(dcache->mappings);
 	free_trans(dcache->mappings);
```
silentbicycle: This doesn't look right.

dhobsd:
yyp is my downfall.

PR #60, opened 2017-08-24
Add tests for the set API

edk0:
Test that the set API does what it's supposed to.

RFC: The build system parts of this are pretty hairy

### DIFF
--- a/tests/set/Makefile
+++ b/tests/set/Makefile
@@ -7,6 +7,7 @@ TEST_OUTDIR.tests/set = ${BUILD}/tests/set
 .for n in ${TEST.tests/set:T:R:C/^set//}
 test:: ${TEST_OUTDIR.tests/set}/res${n}
 SRC += ${TEST_SRCDIR.tests/set}/set${n}.c
+CFLAGS.${TEST_SRCDIR.tests/set}/set${n}.c = -UNDEBUG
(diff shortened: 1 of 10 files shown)


tests/set/Makefile
```diff
@@ -0,0 +1,14 @@
+.include "../../share/mk/top.mk"
```
katef: Grab yourself copyright comments from other files, and stick your name at the top.

tests/set/assert.h
```diff
@@ -0,0 +1,4 @@
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+#include <assert.h>
```
katef: Lose this, and have the makefiles remove -DNDEBUG for the test .c files's `${DFLAGS/CFLAGS.path/to/file.c}` instead.

tests/set/set5.c
```diff
@@ -0,0 +1,33 @@
...
+	int a = *(const int *)a_, b = *(const int *)b_;
+	if (a > b)      return 1;
+	else if (a < b) return -1;
+	else            return 0;
+}
+
+int *next_int(void) {
+	static int n = 0;
+	int *p = malloc(sizeof *p);
+	assert(p);
```
katef: `abort()` for malloc - because it's not what the test is testing for, and because the test still ought to be well-behaved C even if -DNDEBUG is specified.

edk0:
🍰

PR #61, opened 2017-08-31
Add initial property tests for literals, PCRE regexes, sets, priqs, NFAs.

silentbicycle:
This adds a directory with some property-based tests for literals, PCRE dialect regexes,
ADT sets and priqs, and NFAs.

The makefile is based on GNU make, but adapting it to integrate with the overall kmkf
setup shouldn't be much work -- it just compiles and links all files in the directory,
linking them with `libtheft.a`, libfsm, and libre.

Running `build/test_libfsm -l` will list the tests available. `-n` can be used to run
individual tests by a name substring search pattern (e.g. `nfa_`). By default, all run.
`-h` will print other CLI test runner options.

Some of these tests are not expected to pass -- the tests highlight a couple issues,
and will attempt to generate minimal counter-examples to reproduce them.

`main.c`, `test_libfsm.h`, `buf.c`, and `util.c` contain the overall test runner
infrastructure and common code between tests.

`type_info_*.[ch]` files describe how to generate and print instances of various
test input types. For example, `type_info_nfa.c`'s `type_info_nfa_alloc` takes a
theft test runner, then uses its PRNG bitstream to generate a set of
operations building a random NFA.

`test_*.c` files are modules that contain property tests and standalone
regression tests that were previously generated.


fuzz/fuzz_nfa.c
```diff
@@ -553,22 +553,6 @@ nfa_regress_minimise_false_positive(void)
 }
 
 static bool
-segfault_fsm_trim(void)
```
silentbicycle: This should have had "regress_" in the name; it's a regression test for the crash.

fuzz/type_info_adt_set.c
```diff
@@ -57,8 +60,6 @@ alloc_scen(struct theft *t, void *type_env, void **output)
 		}
 	}
 
-	res->count = count;
```
silentbicycle: I wait to set the count until after because (conceptually) until they're ready, the count is zero. In some other generators (such as the NFA one) I have an early exit that drops the overall count, because it gives a better foothold for shrinking.

fuzz/fuzz_nfa.c
katef: I had considered renaming it. But I don't think I want to keep around specially crafted tests for constraints on API contracts, because those ought to come out through ~fuzzing~ property based testing in general.

fuzz/type_info_adt_set.c
katef: Yes, I saw that.

PR #64, opened 2017-09-06
Adds a few niceties to the API

sfstewman:
Adds a function to return the options of an FSM.  Also adds a function that clones an FSM, replacing its options.

### DIFF
--- a/src/libfsm/fsm.c
+++ b/src/libfsm/fsm.c
@@ -82,6 +82,12 @@ fsm_free(struct fsm *fsm)
 	free(fsm);
 }
 
+const struct fsm_options *
+fsm_getoptions(const struct fsm *fsm)
+{
+	return fsm->opt;
+}
+
 void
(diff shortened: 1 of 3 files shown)


src/libfsm/fsm.c
```diff
@@ -82,6 +82,12 @@ fsm_free(struct fsm *fsm)
 	free(fsm);
 }
 
+const struct fsm_options *
+fsm_getoptions(const struct fsm *fsm)
+{
+	return fsm->opt;
+}
+
```
katef: Would `fsm_setoptions()` suit you, after clone?
There are a few operations which call `fsm_new()`, and it'd be odd to provide a `*_with_opts()` variant for all of them.

sfstewman:
Changed approach based on feedback from @katef 

New PR: https://github.com/katef/libfsm/pull/66

Issue #62, opened 2017-09-05
Minimisation by Brzozowski's algorithm produces non-minimal output


katef:
The expected handwritten output tests/minimise/out12.fsm has fewer states than the result produced by Brzozowski's algorithm as implemented by:
```
cat in12.fsm | fsm -rp | fsm -dp | fsm -rp | fsm -dp
```

<img width="330" alt="min12" src="https://user-images.githubusercontent.com/1371085/30069151-c93ce384-9257-11e7-9d11-ad345c754cb1.png">

tests/minimise/in19.fsm shows the handwritten expected minimal output used as input, which remains unchanged.

PR #67, opened 2017-09-13
Exclude NFA start state from its own epsilon closure for reversed FSM with multiple end states

katef:
This addresses the situation described by #62.

### DIFF
--- a/include/fsm/pred.h
+++ b/include/fsm/pred.h
@@ -37,5 +37,11 @@ fsm_hasincoming(const struct fsm *fsm, const struct fsm_state *state);
 int
 fsm_hasoutgoing(const struct fsm *fsm, const struct fsm_state *state);
 
+/*
+ * True iff there are outgoing edges and all are epsilons.
+ */
+int
+fsm_epsilonsonly(const struct fsm *fsm, const struct fsm_state *state);
+
 #endif
 
--- a/src/libfsm/determinise.c
+++ b/src/libfsm/determinise.c
@@ -460,9 +460,34 @@ determinise(struct fsm *nfa,
 	 * This is not yet "done"; it starts off the loop below.
 	 */
 	{
+		const struct fsm_state *nfastart;
 		struct fsm_state *dfastart;
+		int includeself = 1;
 
-		dfastart = state_closure(mappings, dfa, fsm_getstart(nfa), 1);
+		nfastart = fsm_getstart(nfa);
+
+		/*
+		 * As a special case for Brzozowski's algorithm, fsm_determinise() is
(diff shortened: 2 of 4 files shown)


include/fsm/pred.h
```diff
@@ -37,5 +37,11 @@ fsm_hasincoming(const struct fsm *fsm, const struct fsm_state *state);
 int
 fsm_hasoutgoing(const struct fsm *fsm, const struct fsm_state *state);
 
+/*
+ * True iff there are outgoing edges and all are epsilons.
+ */
+int
+fsm_epsilonsonly(const struct fsm *fsm, const struct fsm_state *state);
```
sfstewman: I like the predicate approach for this.

src/libfsm/determinise.c
```diff
@@ -218,6 +219,15 @@ state_closure(struct set *mappings, struct fsm *dfa, const struct fsm_state *nfa
...
+
+		if (set_count(ec) == 0) {
+			set_free(ec);
+			ec = NULL;
+		}
+	}
+
 	if (ec == NULL) {
 		return NULL;
 	}
```
sfstewman: So, given that you call state_closure a lot, and only remove a state once (at the beginning) under very special conditions, does it make sense to pass this as a boolean into the state_closure function?  I'd be tempted to either write a special remove function or to inline the removal.

src/libfsm/determinise.c
```diff
@@ -450,9 +460,34 @@ determinise(struct fsm *nfa,
...
+		 * This is the situation we detect here.
+		 *
+		 * This fabricated start state is then excluded from its epsilon closure,
+		 * so that the closures for its destination states are found equivalent,
+		 * because they also do not include the start state.
+		 *
+		 * If you pass an equivalent NFA where this is not the case (for example,
+		 * with the start state containing an epsilon edge to itself), we regard
+		 * this as any other DFA, and minimal output is not guaranteed.
+		 */
```
sfstewman: The comment looks good to me.  Possibly worth including a reference to the StackExchange answer https://cs.stackexchange.com/questions/6138/problem-with-implementing-brzozowskis-algorithm?noredirect=1&lq=1

src/libfsm/determinise.c
katef: I did wonder that. There's be no other consumer for `state_closure()` with its own state included, so I'd just delete it. And then I got unhappy about losing information. So not today - perhaps when dredging some of this stuff out to a more general interface.

src/libfsm/determinise.c
katef: I saw that! The answer was too terse for me to follow. So I don't think it'd suit this audience, either.

PR #69, opened 2017-09-22
Fixes bug in isdfa query and fsm_isdfa function

sfstewman:
Fixes bug in isdfa predicate and adds a couple of basic tests to detect that isdfa returns true on a DFA and false on an NFA.

See https://github.com/katef/libfsm/commit/ae11b073beaff03327fb06d6fed6d1300f558740 for a description of the bug.


src/libfsm/pred/isdfa.c
```diff
@@ -42,6 +42,8 @@ fsm_isdfa(const struct fsm *fsm, const struct fsm_state *state)
 	 * DFA may not have duplicate edges.
 	 */
 	for (e = set_first(state->edges, &it); e != NULL; e = set_next(&it)) {
+		struct set_iter dst_iter;
+
```
katef: Naming; `jt` by convention for these inner iterators. It rhymes with `it`.

PR #71, opened 2017-09-23
General code to walk two FSMs at once, used to implement intersect and subtract

sfstewman:
Initial version of a suggestion from @katef for a general two-graph walking interface.  The interface is still private to libfsm, but fsm_intersect and fsm_subtract are both reimplemented with it.


katef:
Re-submitted as #79

PR #72, opened 2017-09-23
Skip to avoid numeric values of 0 for set items.

katef:
These are disallowed by the set API since it depends on returning `NULL` for errors.

Here I'm slightly sloppily assuming that `uintptr_t` is larger than `uint16_t`. I didn't feel like introducing a compile-time assertion for that.

### DIFF
--- a/fuzz/fuzz_adt_set.c
+++ b/fuzz/fuzz_adt_set.c
@@ -12,7 +12,7 @@
 struct model {
 	struct set_hook_env *env;
 	size_t bit_count;
-	uint64_t bits[]; /* 0-based */
+	uint64_t bits[];
 };
 
 static enum theft_trial_res prop_set_operations(struct theft *t, void *arg1);
@@ -136,8 +136,11 @@ static enum theft_trial_res prop_set_equality(struct theft *t, void *arg1)
 	set_b = set_create(cmp_item);
 
 	for (i = 0; i < seq->count; i++) {
-		(void) set_add(&set_a, (void *) ((uintptr_t) seq->values[i] + 1U)); /* XXX */
-		(void) set_add(&set_b, (void *) ((uintptr_t) permuted[i]    + 1U));
+		assert(seq->values[i] != 0);
+		assert(permuted[i] != 0);
+
+		(void) set_add(&set_a, (void *) ((uintptr_t) seq->values[i])); /* XXX */
+		(void) set_add(&set_b, (void *) ((uintptr_t) permuted[i]));
 	}
 
 	if (!set_equal(set_a, set_b)) {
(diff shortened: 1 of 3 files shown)


fuzz/type_info_adt_set.h
```diff
@@ -21,10 +21,10 @@ struct set_op {
 	enum set_op_type t;
 	union {
 		struct {
-			uintptr_t item;
+			uintptr_t item; /* 0-based */
```
silentbicycle: Why did you decide to keep the instance 0-based and shift it everywhere it's used? Either approach makes sense, though I probably would have shifted it in the generator instead -- that way I wouldn't have to worry about forgetting to offset it anywhere.

fuzz/fuzz_adt_set.c
```diff
@@ -133,8 +136,8 @@ static enum theft_trial_res prop_set_equality(struct theft *t, void *arg1)
 	set_b = set_create(cmp_item);
 
 	for (i = 0; i < seq->count; i++) {
-		(void) set_add(&set_a, (void *) (uintptr_t) seq->values[i]); /* XXX */
-		(void) set_add(&set_b, (void *) (uintptr_t) permuted[i]);
+		(void) set_add(&set_a, (void *) ((uintptr_t) seq->values[i] + 1U)); /* XXX */
```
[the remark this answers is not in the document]
katef: Those `/* XXX */` are for not handling errors when `set_add()` returns to indicate failure; i'm not doing that in this PR, so they're unchanged here. I'll get them in a bit.

fuzz/type_info_adt_set.h
katef: Because I'm mentally distinguishing the abstract element number (as indexed into the model's bitmap) from the corresponding pointer value. My reasoning is that the only dirty part here is in casting one to the other - so the `(void *)` is the part where I think workarounds belong.

As a thought experiment, I modelled that as indexing into some array of valid non-NULL pointers: `a[item]`. There the rest of the code would be unchanged, and the offset would be "in" each of the pointer values.

`THEFT_TRIAL_SKIP` is a great idea - I'll try that.

fuzz/type_info_adt_set.c
```diff
@@ -45,10 +45,16 @@ alloc_scen(struct theft *t, void *type_env, void **output)
 		switch (op->t) {
 		case SET_OP_ADD:
 			op->u.add.item = theft_random_bits(t, env->item_bits);
+			if (op->u.add.item == 0) {
+				return THEFT_ALLOC_SKIP;
```
silentbicycle: Returning `SKIP` here will leak `res`. It's probably not a big deal, except for extremely long fuzzing runs, but you could call `free(res);` first.

katef: Thanks!

PR #73, opened 2017-09-24
Expose theft's seed to the fuzzing CLI

katef:
Some groundwork here:
* Remove the tests for "slow" situations; I'd like to revisit these in a different way.
* I moved the regression tests to /tests/pcre and /tests/minimise as .re files, because those don't depend on  fuzzing. All that's left under /fuzz now are cases which construct generated input given a seed.
* The seed is passed along from main() to all `test_*()` functions. This means all those share the same seed. I wanted to mention that just to check that the PRNG is re-seeded for each invocation of `theft_run()`.


fuzz/main.c
```diff
@@ -58,11 +63,34 @@ usage(const char *exec_name)
...
 
+static theft_seed
+parse_seed(const char *s)
+{
+	unsigned long u;
+	char *e;
+
+	errno = 0;
+
+	u = strtoul(s, &e, 16);
```
silentbicycle: Does this need to be `strtoull`? The seed is a typedef of `uint64_t`, so it could get truncated as a long.

fuzz/fuzz_adt_priq.c
```diff
@@ -120,7 +120,7 @@ test_priq_operations(uintptr_t limit)
 			.enable = true,
 		},
 
-		.seed = theft_seed_of_time(),
+		.seed = seed,
```
silentbicycle: > The seed is passed along from main() to all test_*() functions. This means all those share the same seed. I wanted to mention that just to check that the PRNG is re-seeded for each invocation of theft_run().

Correct - if you set the seed here, it will use that seed for the _first_ trial, but subsequent trials will use different seeds derived from that seed somehow. IIRC, it re-seeds the PRNG with the last trial seed and draws a new random value for the next, but it could also just increment the seed or something. It's intentionally an implementation detail how seeds are chosen, besides the initial seed and "always" seeds, and may change in a future version.

The PRNG is currently not seeded to a consistent value during the property test function itself (because a `struct theft *t` handle is passed in, you can draw new random values during the property test, in case probabilistic algorithms are being used), but that's something I plan on fixing in the next release.

fuzz/fuzz_adt_set.c
```diff
@@ -68,7 +68,7 @@ test_set_operations(uintptr_t limit2)
 			.trial_post = repeat_with_verbose,
 			.env = &env,
 		},
-		.seed = theft_seed_of_time(),
+		.seed = seed,
 		.trials = 100000,
```
silentbicycle: Out of scope of this PR, but the trial count might be another good CLI argument. At that point, it might also be better to pass in a config struct to the test (with the seed, trial count, and whatever else).

katef: I was thinking just the same - but I'd like to be able to specify that in terms of something meaningful for each particular test. Because they have different densities for their search spaces. I feel like such a CLI argument ought to apply equivalently for all tests. Thoughts?

silentbicycle: They should probably have different defaults, agreed. If I'm overriding the trial count, I'm usually also just running a single test by name, though. Running all of the properties (either from CI or directly) is usually part of a different workflow for me than running a single property with a huge number of trials (possibly fuzzing it for several hours).

PR #74, opened 2017-10-03
Add some PCRE tests

edk0:


tests/minimise/in17.re
```diff
@@ -0,0 +1 @@
+[aBx]*a
```
katef: We already have this as `in0.re`

edk0: I'm not sure how this ended up here; removed.

katef: It's getting to be a famous test case now :)

tests/pcre/in16.re
```diff
@@ -0,0 +1 @@
+[[:alpha:]]
```
katef:
```
build/bin/re -r pcre -py tests/pcre/in16.re  > build/tests/pcre/got16.fsm
tests/pcre/in16.re:1: Syntax error: expected group body
```
what's going on here?

edk0: That's the effect of pre-#75. I'll rebase so the PR head builds.

katef: Oh of course. silly me.

PR #76, opened 2017-10-09
Fixes src/libfsm/libfsm.syms for OS X

sfstewman:
Small problem with a comment in the symbol listing.  OS X doesn't
support comments prefixed with '#', and the sed command used to remove
the comments only works with comments at the beginning of the line.

Changed one of the comments so it starts at the beginning of the line.

### DIFF
--- a/src/libfsm/libfsm.syms
+++ b/src/libfsm/libfsm.syms
@@ -84,4 +84,5 @@ fsm_exec
 fsm_fgetc
 fsm_sgetc
 
-fsm_out_c # XXX: workaround for lx
+# XXX: workaround for lx
+fsm_out_c


katef:
Thanks!
There's no comment stripping - the sed command you're thinking of (in kmkf's part.mk) prefixes an `_` for mach-o symbols. I think I should add comment stripping, too.

katef:
I did that here: https://github.com/katef/kmkf/commit/2f0c4e1858319fcd0b18572fb6295c92d475d8b4

katef:
Added locally in 2f0c4e1858319fcd0b18572fb6295c92d475d8b4

PR #77, opened 2017-10-10
General function to walk two FSMs at once

sfstewman:
Initial version of a suggestion from @katef for a general two-graph walking interface. The interface is still private to libfsm.  fsm_intersect and fsm_subtract can both be implemented by calling it with different options.


src/libfsm/intersect.c
```diff
@@ -6,13 +6,15 @@
...
-#include <limits.h>
 #include <errno.h>
 
 #include <fsm/fsm.h>
+#include <fsm/out.h>
+#include <fsm/options.h>
 #include <fsm/bool.h>
 
 #include "internal.h"
+#include "walk2.h"
```
katef: I'd rather no change to this file on this PR; same for subtract.c.

src/libfsm/walk2.c
```diff
@@ -0,0 +1,469 @@
+#include "walk2.h"
```
katef: Copyright

src/libfsm/walk2.c
```diff
@@ -0,0 +1,469 @@
...
+#include <adt/set.h>
+
+#include <fsm/fsm.h>
+#include <fsm/out.h>
+#include <fsm/options.h>
+#include <fsm/bool.h>
+#include <fsm/pred.h>
+#include <fsm/walk.h>
+
+#include "internal.h"
```
katef: "walk2.h" goes here, since it's internal

src/libfsm/walk2.c
```diff
@@ -0,0 +1,469 @@
...
+
+	delta = (pa->a > pb->a) - (pa->a < pb->a);
+	if (delta == 0) {
+		delta = (pa->b > pb->b) - (pa->b < pb->b);
+	}
+
+	return delta;
+}
+
+struct fsm_walk2_tuple_pool {
```
katef: Better explain what this is, since it's not part of the algorithm proper.

src/libfsm/walk2.c
```diff
@@ -0,0 +1,469 @@
...
+
+	item = &pool->items[0];
+	*item = zero;
+	return item;
+}
+
+static unsigned
+walk2mask(int has_a, int has_b)
+{
+	int endbit = (!!has_a << 1) | !!has_b;
```
katef: Is this reproducing the same bits as the enum provides?

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
...
+
+struct fsm_walk2_tuple *
+fsm_walk2_tuple_new(struct fsm_walk2_data *data, struct fsm_state *a, struct fsm_state *b);
+
+void
+fsm_walk2_data_free(struct fsm_walk2_data *data);
+
+/* Sets all of the equiv members of the states of the fsm to NULL */
+void
+fsm_walk2_mark_equiv_null(struct fsm *fsm);
```
katef: Can this be static?

src/libfsm/walk2.c
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	/* mark combined state as visited */
+	qc->equiv = qc;
+
+	/* fsm_walk2_edges walks the edges of two graphs, generating combined
+	 * states.
+	 *
+	 * This is a synthesis of two separate walk functions, one for
+	 * intersecting graphs and one for subtracting them.
```
katef: That seems like history; I don't think it's useful to the reader here.

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
+#ifndef FSM_WALK2_H
```
katef: Copyright

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
...
+#define FSM_WALK2_H
+
+#include <stdlib.h>
```
katef: Not here please

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
...
+};
+
+/* Constraints for walking edges for union (A|B), intersection (A&B), and
+ * subtraction (A-B)
+ */
+enum {
+	FSM_WALK2_EDGE_UNION     = FSM_WALK2_ONLYA | FSM_WALK2_ONLYB | FSM_WALK2_BOTH,
+	FSM_WALK2_EDGE_INTERSECT = FSM_WALK2_BOTH,
+	FSM_WALK2_EDGE_SUBTRACT  = FSM_WALK2_ONLYA | FSM_WALK2_BOTH
+};
```
katef: I'd rather not provide these; I think use-cases aren't part of walk2's API.

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Notice that the edge constraint for subtraction is different from the
+ * end state constraint.  With subtraction, you want to follow edges
+ * that are either ONLYA or BOTH, but valid end states must be ONLYA.
+ */
+enum {
+	FSM_WALK2_END_UNION     = FSM_WALK2_ONLYA | FSM_WALK2_ONLYB | FSM_WALK2_BOTH,
+	FSM_WALK2_END_INTERSECT = FSM_WALK2_BOTH,
+	FSM_WALK2_END_SUBTRACT  = FSM_WALK2_ONLYA
+};
```
katef: Likewise
(and the comment about the constraint for subtraction would then go in subtract.c)

src/libfsm/walk2.h
```diff
@@ -0,0 +1,109 @@
...
+	 * is a valid end state or not.  We need four bits.
+	 *
+	 * Operation	Requirement			endmask
+	 * intersect    both true			0x8
+	 * subtract     first true, second false	0x4
+	 * union	either true			0xE
+	 */
+	unsigned endmask:4;  /* bit table for what states are end states in the combined graph */
+	unsigned edgemask:4; /* bit table for which edges should be followed */
+};
```
katef: Can these structs be private to walk2.c?

src/libfsm/walk2.c
sfstewman: I had that originally, but it's too easy to mask include dependencies.  This guarantees a compiler error if walk2.h requires includes that it doesn't provide.

src/libfsm/walk2.h
sfstewman: walk2.h requires size_t to be defined

src/libfsm/walk2.h
sfstewman: Would it make sense to keep everything private except the fsm_walk2 call and the relevant enums?

src/libfsm/intersect.c
sfstewman: The intermediate changesets involve quite a few changes to intersect.c.  Should I squash them into one changeset or are intermediates okay as long as the PR as a whole leaves intersect.c (and others) untouched?

katef: Whatever works for you. I think if it were me, I'd remove everything from the history which edited this file (by `git rebase -i` and editing/dropping those commits). But I don't mind.

src/libfsm/walk2.c
katef: Yes, but it goes here.

src/libfsm/walk2.c
katef: Okay. Then I don't like that the knowledge about bits is repeated in those two places. Could you phrase this in terms of the enum's constants perhaps?

src/libfsm/walk2.h
katef: Then the caller will provide it.

src/libfsm/walk2.h
katef: That's what I'm asking - if I understand the interface correctly, callers don't need to know about anything except for that prototype and the one enum.

src/libfsm/intersect.c
sfstewman: The changes are quite extensive; fsm_walk2 grew out of the intersection-by-dfs code in intersect.c

I should probably squash the commits so it's more sensible to a reader that doesn't know the sordid history. :)

src/libfsm/walk2.c
sfstewman: I can't think of a way to do this without a lookup table, several ?: expressions, or a bunch of if statements.  Seems like a waste to introduce a 2x2 lookup table here, and the other options add some ugly branching.

src/libfsm/walk2.h
sfstewman: I think that's correct.  I'll move the rest inside of walk2.c

src/libfsm/walk2.c
katef: I was picturing if statements. or `switch ((has_a << 1) | has_b)`. The point being that that's in terms of the local variables here, and doesn't make any assumptions about the bit patterns for the enum constants.

katef:
Re-submitted as #79

PR #86, opened 2018-06-10
lx: Retain/restore column position around newlines in `*getc/*ungetc`.

silentbicycle:
This adds a field to `struct lx_pos`, `.saved_col`, which is used to
save and restore the column position when `getc`/`ungetc` handle '\n'.

If I'm understanding the test setup correctly, this change does not
cause any new test failures. I have a small standalone test program
which uses an lx-genearted lexer with `-b dyn -k str`, sinks various
known strings into it, and checks that the line/colum/byte positions are
as expected. I could add that to the PR, but I'm not sure how to
integrate a standalone .c test program into the build/test config.

This implementation wastes an unsigned (`.saved_col` is unused in
`lx->start`), which could be avoided, but would likely complicate
the other code generation.

### DIFF
--- a/src/lx/out/c.c
+++ b/src/lx/out/c.c
@@ -478,7 +478,8 @@ out_io(FILE *f)
 		fprintf(f, "\n");
 		fprintf(f, "\tif (c == '\\n') {\n");
 		fprintf(f, "\t\tlx->end.line++;\n");
-		fprintf(f, "\t\tlx->end.col = 1;\n");
+		fprintf(f, "\t\tlx->end.saved_col = lx->end.col - 1;\n");
+		fprintf(f, "\t\tlx->end.col = 0;\n");
 		fprintf(f, "\t}\n");
 		fprintf(f, "\n");
 	}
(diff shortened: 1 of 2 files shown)


src/lx/out/c.c
```diff
@@ -526,7 +527,7 @@ out_io(FILE *f)
 		fprintf(f, "\n");
 		fprintf(f, "\tif (c == '\\n') {\n");
 		fprintf(f, "\t\tlx->end.line--;\n");
-		fprintf(f, "\t\tlx->end.col = 0; /* XXX: lost information */\n");
+		fprintf(f, "\t\tlx->end.col = lx->end.saved_col;\n");
```
katef: That's a surprisingly simple solution. It hadn't occurred to me at all!

katef:
> I have a small standalone test program which uses an lx-genearted lexer with -b dyn -k str, sinks various known strings into it, and checks that the line/colum/byte positions are as expected.
There are generated tests for the various `-b` implementations; if you build with `bmake -r DEBUG=1 CC=gcc test` you should see them under `build/tests/dump-lx`, with the output to *.tok files there.

I think it'd make sense to add the position values to the dumped output there; `lx -l dump` does include it already, although only as a byte offset (see `API_POS` in src/lx/out/dump.c). On second thoughts, adding that wouldn't ensure that the line & column are correct, just that they correspond for each `-b` implementation.

If you want to add your standalone test under tests/, I'll do the makefile glue.

src/lx/out/c.c
silentbicycle: I'm not sure it's right, it seems to give the wrong result when the input string has newlines, like `x\ny\nz\n`.

silentbicycle:
Merged in master, no conflicts; this should be good to go.

PR #89, opened 2018-06-25
Add ASTs for libre dialects

silentbicycle:
Add abstract syntax trees for the various dialects, so compilation can make decisions based on the entire parsed regex rather than building an FSM at parse time.

- Remove some unnecessary states and epsilon transitions from the regexs' NFAs, since after parsing, it's clear they are no longer necessary. While minimization would remove them anyway, it should make other processing more efficient.

- Add prettyprinting for the parsed AST. Rather than speculate too much about what output format might be useful (JSON? dot?), the current format is pretty basic, mostly to aid development.

- Add some logging during AST construction and compilation. (Note that in its current form, this adds a couple warnings for variadic macros being used in an otherwise `-std=c89` codebase.)

- Switch from distinct `CLASS_*` named character class token types to a single type, `NAMED_CHAR_CLASS`. This significantly reduces the surface area for the grammar and semantic actions, and will make it significantly easier to implement PCRE negated named character classes or other new character classes later.

- Add a new error type, `<err-unsupported>`, for constructs (mostly in PCRE) that are detected at parse time but explicitly unsupported.

- Add a distinct token type for named character classes.

- Fix some bugs related to cases in PCRE where named character classes and character types have different rules.

- Add PCRE character class ranges with character types and/or named character classes as well as character literals.

- Correct some subtleties regarding PCRE's escaped character handling rules.

- Add a link to the PCRE reference in its grammar.

- Document part of the grammar that seems to require LL(2).

- Fix pcre \Qquoting\E.

- Add lots of tests, including some tests for expected error messages given for invalid regexes.

Note: Some of the build config needs to be updated -- rather than using the overarching makefile setup to run tests, I've been using the `testdir` script, to focus on a particular dialect at a time.

With these changes, all of the dialect-related tests are passing except for:

    - tests/pcre/in10.re: `\c` is explicitly unsupported, so this won't work
    - tests/pcre/in14.re: This doesn't seem to detect that `\x00-\xff` can be replaced with `.`
    - tests/native/in8.re: ^ and $ anchors aren't implemented yet


src/libre/re_char_class.c
```diff
@@ -0,0 +1,802 @@
...
+	assert(cc != NULL);
+
+	filter = char_filter_for_class_id(id);
+
+	if (filter == NULL) {
+		return 0;
+	}
+
+	for (i = 0; i < 256; i++) {
+		if (filter(i)) { cc_add_byte(cc, i); }
```
katef: I think iterating 0..UCHAR_MAX here is the wrong approach; the named classes are provided as DFA currently because their contents need to handle both UTF-8 sequences and ASCII.

Perhaps the AST ought to think in terms of codepoints, rather than bytes. I'm not sure.

My instinct is that the "filter" functions should not exist, because we shouldn't be running things in order to construct them. However we express them, I'd hope these sets can be constructed ahead of time.

src/libre/re_char_class.c
```diff
@@ -0,0 +1,802 @@
...
+	
+	return 1;
+}
+
+void
+cc_invert(struct re_char_class *cc)
+{
+	unsigned i;
+	for (i = 0; i < sizeof(cc->chars)/sizeof(cc->chars[0]); i++) {
+		cc->chars[i] = ~cc->chars[i];
```
katef: Not for unicode!

src/libre/re_comp.c
```diff
@@ -0,0 +1,320 @@
...
+		break;
+	}
+
+	case AST_EXPR_CHAR_TYPE:
+	{
+		int i;
+		struct re_char_class *cc = re_char_class_type_compile(n->u.char_type.id);
+		if (cc == NULL) { return 0; }
+
+		for (i = 0; i < 256; i++) {
```
katef: Any time you're iterating to 0..256, please make that `UCHAR_MAX` instead; it's something I grep for when considering character set support, and what can be bitmaps, and so on.

src/libre/re_comp.c
```diff
@@ -0,0 +1,320 @@
...
+	env.opt = opt;
+	env.flags = flags;
+	
+	x = fsm_getstart(env.fsm);
+	assert(x != NULL);
+	
+	y = fsm_addstate(env.fsm);
+	if (y == NULL) {
+		goto error;
+	}
```
katef: Lovely to see these pulled out of `parser.act`

src/libre/re_ast.c
```diff
@@ -0,0 +1,389 @@
...
+
+#define INDENT(F, IND)							\
+	do {								\
+		size_t i;						\
+		for (i = 0; i < IND; i++) { fprintf(F, " "); }		\
+	} while(0) 							\
+
+static void
+pp_iter(FILE *f, size_t indent, struct ast_expr *n)
+{
```
katef: Could you move the AST (and character class) pretty-printing to its own file please? At some point we can pull that out to a subdirectory and provide various output formats. I'm after dot and some kind of BNF especially.

src/libre/re_char_class.c
```diff
@@ -0,0 +1,802 @@
...
+	return res;
+}
+
+
+static void
+bitset_pos(unsigned char byte, unsigned *pos, unsigned char *bit)
+{
+	*pos = byte / 8;
+	*bit = 1U << (byte & 0x07);
+}
```
katef: I forgot to say, there's a bitmap under `<adt/bitmap.h>` already - which the character classes did originally use before converting them to graft on pre-constructed DFA as part of the unicode support.

include/re/re.h
```diff
@@ -112,9 +114,14 @@ struct re_err {
 int
 re_flags(const char *s, enum re_flags *f);
 
+/* Callback to acquire each character of the input, in the spirit of fgetc().
+ * opaque can be used to pass in callback-specific state, and is otherwise
+ * opaque to libre. */
+typedef int
+re_getchar_fun(void *opaque);
+
```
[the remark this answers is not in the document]
silentbicycle: Agreed. Should that happen in this PR or another? I don't think anything else in this PR touches libfsm at all.

src/libre/dialect/literal/parser.sid
```diff
@@ -67,81 +54,93 @@
...
 	};
 
-	re_literal: (x :fsm_state, y :fsm_state) -> () = {
+	re_literal: () -> (n :ast_expr) = {
 		{
-			list-of-literals(x, y);
-		||
-			<add-epsilon>(x, y);
+			n = list-of-nodes();
 		##
```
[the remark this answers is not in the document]
silentbicycle: You're right, it's a legal parse and should use `||`. Fixed in 017098c (for all the dialects).

src/libre/dialect/pcre/lexer.lx
```diff
@@ -37,28 +37,26 @@
...
-'\d' -> $class_digit;
-'\D' -> $class_ndigit;
-'\w' -> $class_word;
-'\W' -> $class_nword;
-'\s' -> $class_space;
-'\S' -> $class_nspace;
+# character type escapes
+# Note: these need to be their own lexemes distinct from named character
+# classes such as /[:digit:]/, as they are allowed in different contexts
+char_type = '\' /[dDhHsSvVwWN]/;
```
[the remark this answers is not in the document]
silentbicycle: Unfortunately yes, it breaks the symmetry. I believe I have a note to that effect in the header.

`\N` is supposed to be the opposite of `\n`, since (unless a PCRE flag is set so `.` also matches across lines, off by default) it matches everything except newlines.

The individual `$char__type` patterns are separated in e0561b9.

src/libre/dialect/pcre/parser.sid
```diff
@@ -4,17 +4,26 @@
  * See LICENCE for the full copyright terms.
  */
 
+/* Regular expressions, as evaluated by PCRE, which supports the
+ * patterns listed at:
+ *     http://pcre.org/current/doc/html/pcre2pattern.html
+ *
+ * Note that not all regex extensions are supported yet. */
```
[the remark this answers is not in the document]
silentbicycle: Noted in 1a3cccf.

src/libre/dialect/pcre/lexer.lx
```diff
@@ -81,44 +79,49 @@ class = '\' /[dDwWsS]/;
...
-	'[:digit:]'  -> $class_digit;
-	'[:graph:]'  -> $class_graph;
-	'[:lower:]'  -> $class_lower;
-	'[:print:]'  -> $class_print;
-	'[:punct:]'  -> $class_punct;
-	'[:space:]'  -> $class_space;
-	'[:upper:]'  -> $class_upper;
-	'[:word:]'   -> $class_word;
-	'[:xdigit:]' -> $class_xdigit;
+        # FIXME: negated character classes, e.g. [:^alpha:]
```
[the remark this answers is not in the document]
silentbicycle: If the name -> `enum ast_char_class_id` mapping is moved into a dialect-specific function (which would also be my preference), we could either have extra enum values there, or also use a `negated` flag like the char_type matching does.

src/libre/parser.act
```diff
@@ -77,6 +78,15 @@
 	#define TOK_CLASS__nspace  TOK_CLASS_NSPACE
 	#define TOK_CLASS__ndigit  TOK_CLASS_NDIGIT
 
+	/* This is a hack to work around the AST files not being able to include lexer.h. */
```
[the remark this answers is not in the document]
silentbicycle: Because that type, `struct ast_pos`, is defined in `re_char_class.h` and used in `re_char_class.c` (as well as the `.act` file), which is compiled once rather than per-dialect, so it doesn't have access to the dialect-specific details. The `lx_pos` struct it's duplicating is the same for every dialect, but that isn't known when building that .c file.

src/libre/parser.act
```diff
@@ -1299,44 +1123,31 @@
...
+	DIALECT_PARSE(re_getchar_fun *f, void *opaque,
 		const struct fsm_options *opt,
 		enum re_flags flags, int overlap,
 		struct re_err *err)
 	{
-		struct fsm *new;
-		struct fsm_state *x, *y;
+		/* struct fsm *new; */
+		struct ast_re *ast;
+		/* struct fsm_state *x, *y; */
```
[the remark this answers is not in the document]
silentbicycle: Oops. Fixed in bd84be7.

src/libre/re.c
```diff
@@ -103,11 +105,27 @@ re_comp(enum re_dialect dialect, int (*getc)(void *opaque), void *opaque,
...
 	}
 
-	new = m->comp(getc, opaque, opt, flags, m->overlap, err);
-	if (new == NULL) {
+	ast = m->parse(getc, opaque, opt, flags, m->overlap, err);
+	if (ast == NULL) {
 		return NULL;
 	}
 
+	/* Do a complete pass over the AST, filling in other details. */
```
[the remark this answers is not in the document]
silentbicycle: Now that parse has finished and the AST is complete, we can do a full top-down sweep over the entire AST for other details that difficult to decide locally, at parse time.

Currently, this pass just numbers the match group nodes, but soon it will also set info about ^ and $ anchoring (which depends on anchoring details for all alts in any `(x|y|z)` group), and could also do other rewriting.

src/libre/re_ast.c
silentbicycle: Moved in ae2f91d.

src/libre/dialect/pcre/parser.sid
```diff
@@ -60,417 +51,379 @@
...
-				<count-m-to-n>(x, &y, n, n);
+				c = <expr-count>(x, start, x, end);
+				n = <ast-expr-atom>(e, c);
 			||
 				(start, !) = OPENCOUNT;
-				m = COUNT;
+				x = COUNT;
 				SEP;
-				n = COUNT;
+				y = COUNT;
```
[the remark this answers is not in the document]
silentbicycle: Fixed in 23a0958 (along with renaming the `n` AST return value to `node` throughout).

include/re/re.h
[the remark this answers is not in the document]
katef: I don't mind really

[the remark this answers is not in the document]
silentbicycle: I'd rather it happen in another PR, this one is already huge.

silentbicycle:
Character types have been integrated into character classes, and character classes are now based on the `class.h` FSMs. There are a couple that are currently unsupported (horiz/vert whitespace and negated ones like `\W`) .

SQL subtraction of character classes is currently broken (`/[a-z^aeiou]/`) but that isn't implemented on master either -- it could be fixed, or flagged as unsupported. Thoughts?

silentbicycle:
`/[a-z^aeiou]/` is now detected and returns unsupported, as does using any unsupported character classes or types. The missing character classes can probably be added in a different PR, I think otherwise this is ready for further review.

src/libre/dialect/pcre/re_dialect_pcre.c
```diff
@@ -0,0 +1,73 @@
...
+	for (i = 0; t && t[i].s != NULL; i++) {
+		if (0 == strncmp(t[i].s, name, strlen(t[i].s))) {
+			if (t[i].ctor == NULL) { return RE_CLASS_UNSUPPORTED; }
+			*res = t[i].ctor;
+			return RE_CLASS_FOUND;
+		}
+	}	
+
+	return RE_CLASS_NOT_FOUND;
+}
```
katef: The dialect-specific `re_char_class_*()` functions look great; all the concepts there seem like everything's in the right place and nothing's exposed where it shouldn't be.

Using the `char_class_constructor_fun` function pointer as essentially an enum is kind of odd, but it seems to work out!

src/libre/re_analysis.c
```diff
@@ -0,0 +1,77 @@
...
+		break;
+	case AST_EXPR_CHAR_CLASS:
+		analysis_char_class(env, n->u.char_class.cca);
+		break;
+	case AST_EXPR_GROUP:
+		/* assign group ID */
+		env->group_id++;
+		n->u.group.id = env->group_id;
+		analysis_iter(env, n->u.group.e);
+		break;
```
katef: There are some things that the `re_ast_analysis()` pass is doing which could definitely be done on-the-fly during parse, but i know there's more coming for anchors there, and i'm happy to leave that as-is and think about it again after things settle down.

For example, i think group numbering would be pretty natural to supply with a SID action keeping the current maximum in `struct act_state` and providing a sequential ID each time it's called.

src/libre/re_ast.h
```diff
@@ -0,0 +1,145 @@
...
+enum ast_expr_type {
+	AST_EXPR_EMPTY,
+	AST_EXPR_CONCAT,
+	AST_EXPR_ALT,
+	AST_EXPR_LITERAL,
+	AST_EXPR_ANY,
+	AST_EXPR_MANY,
+	AST_EXPR_KLEENE,
+	AST_EXPR_PLUS,
+	AST_EXPR_OPT,
```
katef: `AST_EXPR_KLEENE` and friends still exist in `enum ast_expr_type`

src/libre/re_char_class.h
```diff
@@ -0,0 +1,182 @@
...
+		} named;
+		struct {
+			enum re_char_class_flags f;
+		} flags;
+		struct {
+			struct re_char_class_ast *ast;
+			struct re_char_class_ast *mask;
+		} subtract;
+	} u;
+};
```
katef: Probably also after this settles, I would like to consider making `struct re_char_class_ast` less tree-shaped - I think we can get rid of the `.concat` node and have classes as an array of `enum re_char_class_ast_type` items.

src/libre/parser.act
```diff
@@ -31,6 +30,8 @@
 	#include <ctype.h>
 
 	#include "libfsm/internal.h" /* XXX */
+	#include "libre/re_ast.h"
+	#include "libre/re_char_class.h"
 
 	#include <fsm/fsm.h>
 	#include <fsm/bool.h>
```
katef: Can `<fsm/*.h>` and `"libfsm/internal.h"` go now?

src/libre/parser.act
```diff
@@ -102,6 +115,14 @@
 	typedef struct fsm * t_fsm;
```
katef: Can `t_fsm` go too? and perhaps `struct group` and `t_grp`.

katef: I think we should be left with no FSM stuff in the parsers.

src/libre/dialect/sql/parser.sid
```diff
@@ -75,260 +63,265 @@
...
-*/
-				$;
+				f = <char-class-ast-flag-none>();
 			};
-		};
 
-		group-bm: (g :grp &) -> () = {
-			(start, !) = OPENGROUP;
+			char-class-tail: () -> (f :char_class_ast) = {
+			/* This requires LL(2). */
```
katef: SID does actually have a mechanism for something like this - you can supply a predicate function, written in C, which can run whatever you like. The result influences which alt to take. It's used in TenDRA for parts of C++'s grammar which are not resolvable without consulting a symbol table, or whatever else. (And also for the typedef problem in C, I think). It really suits ad-hoc lookahead. Now's not the time for it though.

src/libre/re_analysis.c
silentbicycle: The group anchoring could definitely move into the parser, but it was a simple case to prove out the top-down analysis pass.

As far as I can tell, a separate analysis pass is either necessary to make anchoring work properly, or will greatly simplify the implementation.

src/libre/parser.act
silentbicycle: Yes they can! Removed in 07127fa.

src/libre/parser.act
silentbicycle: Indeed. Removed in f5adce5 .

PR #91, opened 2018-07-02
AST rendering to multiple outputs

katef:
This PR introduces `src/libre/print/*` for printing to various formats, exposed alongside the existing `re -l xyz` names.

These are:
* `re -pl tree` to print the existing text dump;
* `re -pl ast` for graphviz;
* `re -pl ebnf` for ISO EBNF, suitable for feeding to kgt;
* `re -pl pcre` to render to PCRE regexp syntax

None of these are particularly capable yet; I just wanted to get them in place for sake of debugging, and settling out ideas.

You can render with kgt like so:
```
; ./build/bin/re -pl ebnf 'a{3}b*[abc]|xyz' | kgt -l iso-ebnf -e rrtext
e:
    ||--v-->---- "a" ----v-->---------v--v-- "a" -->-->--||
        |  |             |  |         |  |         |  |
        |  ^--(2 times)--<  ^-- "b" --<  >-- "b" --^  |
        |                                |         |  |
        |                                >-- "c" --^  |
        |                                             |
        >------------- "x" -- "y" -- "z" -------------^
```

### DIFF
--- a/src/libre/print.h
+++ b/src/libre/print.h
@@ -19,6 +19,9 @@ void
 re_ast_print_dot(FILE *f, const struct fsm_options *opt, struct ast_re *ast);
 
 void
+re_ast_print_pcre(FILE *f, const struct fsm_options *opt, struct ast_re *ast);
+
+void
 re_ast_print_tree(FILE *f, struct ast_re *ast);
 
 #endif
--- a/src/libre/print.h
+++ b/src/libre/print.h
@@ -16,6 +16,9 @@
 #define PRETTYPRINT_AST 0
 
 void
+re_ast_print_dot(FILE *f, const struct fsm_options *opt, struct ast_re *ast);
+
+void
 re_ast_print_tree(FILE *f, struct ast_re *ast);
 
 #endif
--- a/src/libre/print.h
+++ b/src/libre/print.h
@@ -16,6 +16,7 @@ typedef void (re_ast_print)(FILE *f, const struct fsm_options *opt,
 	const struct ast_re *ast);
 
 re_ast_print re_ast_print_dot;
+re_ast_print re_ast_print_ebnf;
 re_ast_print re_ast_print_pcre;
 re_ast_print re_ast_print_tree;
 
--- a/src/libre/print.h
+++ b/src/libre/print.h
@@ -7,21 +7,18 @@
 #ifndef RE_INTERNAL_PRINT_H
 #define RE_INTERNAL_PRINT_H
 
-#include <stdlib.h>
 #include <stdio.h>
-#include <assert.h>
-
-#include "re_ast.h"
 
 #define PRETTYPRINT_AST 0
 
-void
-re_ast_print_dot(FILE *f, const struct fsm_options *opt, struct ast_re *ast);
+struct ast_re;
+struct fsm_options;
 
-void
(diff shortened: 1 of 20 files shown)


src/libre/print.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifndef RE_INTERNAL_PRINT_H
+#define RE_INTERNAL_PRINT_H
+
+#include <stdio.h>
+
+struct ast_re;
+struct fsm_options;
+
+typedef void (re_ast_print)(FILE *f, const struct fsm_options *opt,
```
silentbicycle: The parens around the typedef'd function name aren't strictly necessary.

src/libre/print/dot.c
```diff
@@ -0,0 +1,222 @@
...
+}
+
+static void
+pp_iter(FILE *f, const struct fsm_options *opt,
+	const void *parent, struct ast_expr *n)
+{
+	assert(f != NULL);
+	assert(opt != NULL);
+
+	if (n == NULL) { return; }
```
silentbicycle: We should soon be able to assert on `n` never being NULL -- the branch with anchoring will also thoroughly switch to using a single interned `EMPTY` node as a sentinel, so a NULL pointer would be an internal error.

src/libre/print/ebnf.c
```diff
@@ -0,0 +1,294 @@
...
+		return 1;
+
+	case AST_EXPR_REPEATED:
+	case AST_EXPR_CONCAT:
+	case AST_EXPR_ALT:
+	case AST_EXPR_MANY:
+		return 0;
+
+	case AST_EXPR_FLAGS:
+		return 0; /* XXX */
```
silentbicycle: Does the `XXX` here indicate you still mean to change this? `0` seems like a reasonable return value.
I think of the flag nodes as being purely metadata. I might be able to do away with them entirely, once the anchoring is fully working, because there's also node flags used for tracking the anchoring-related state.

src/libre/print/ebnf.c
```diff
@@ -0,0 +1,294 @@
...
+	fprintf(f, "e = ");
+
+	pp_iter(f, opt, ast->expr);
+
+	fprintf(f, "\n");
+	fprintf(f, "  ;");
+	fprintf(f, "\n");
+}
+
+static void
```
silentbicycle: Some of these are duplicated in every output format, would it be worth providing a common default but being able to override them?

src/libre/print/pcre.c
```diff
@@ -0,0 +1,247 @@
...
+	case AST_EXPR_REPEATED: {
+		size_t i;
+
+		static const struct {
+			unsigned m;
+			unsigned n;
+			const char *op;
+		} a[] = {
+#define _ AST_COUNT_UNBOUNDED
+			{ 0, 0, NULL },
```
silentbicycle: Nice.

src/libre/print/ebnf.c
katef: It means I question the line - in this case I think `0` probably makes sense, but something feels wrong. I don't know what.

src/libre/print/ebnf.c
katef: I'd rather come back and refactor them all together. I didn't want to touch tree.c in particular, while you're nearby.

PR #94, opened 2018-07-13
Introduce codegen IR

katef:
This PR introduces an IR datastructure for libfsm after the DFA and before the code generation output.

The main idea is that the decisions for code generation are done here, rather than when printing code. This leaves the code generation to simply walk the IR, printing what's there, without mixing in logic. There are still a few choices which might be made during the printing of code, but those ought to be cosmetic only.

This IR is intended for producing code only. Some of the output from libfsm expresses state machines verbatim (especially libfsm/print/dot.c which attempts to show an FSM compactly but verbatim). Those still walk the FSM directly.

IR nodes correspond approximately to states in the FSM. This may change over time when various transformations possibly combine IR nodes.

There are several node types, for the various "strategies" of generating code for a particular state. Here's an example showing most of them:
```
; ./build/bin/re -pl ir -z '[^a-z].' 'm+[0-9a-f]([^a-zA-Z]A|[a-z]B|.B)?' \
    | dot -Tpng -o /tmp/x.png
```
IR
and its corresponding DFA:
DFA


src/libfsm/print/ir.h
```diff
@@ -0,0 +1,106 @@
...
+ * The fsm_state structs don't exist from this IR onwards.
+ * These nodes are always DFA. No epsilons, No duplicate edge labels,
+ * so this IR takes advantage of that.
+ *
+ * But this is not neccessarily a minimal DFA. By this stage,
+ * we're done with graph algorithmics.
+ */
+
+enum ir_strategy {
+	IR_NONE     = 1 << 0,
```
silentbicycle: Based on the comment below, is this simultaneously being used as a type tag for the `struct ir_state` union, but also as a set of allowed strategies in `make_ir`?

src/libfsm/print/ir.h
```diff
@@ -0,0 +1,106 @@
...
+	IR_SAME     = 1 << 1,
+	IR_COMPLETE = 1 << 2,
+	IR_PARTIAL  = 1 << 3,
+	IR_DOMINANT = 1 << 4,
+	IR_ERROR    = 1 << 5,
+	IR_TABLE    = 1 << 6
+};
+
+struct ir_range {
+	unsigned char start;
```
silentbicycle: Is it worth noting that this is inclusive? Is this something that will need to be adapted for UTF-8?

src/libfsm/print/ir.h
```diff
@@ -0,0 +1,106 @@
...
+};
+
+struct ir {
+	size_t n;
+	unsigned start;
+	struct ir_state *states; /* array */
+};
+
+/* TODO: can pass in mask of allowed strategies */
+struct ir *
```
silentbicycle: Should these be prefixed/namespaced, if they're part of the symbols exported in libfsm?

src/lx/print/c.c
```diff
@@ -19,15 +19,16 @@
...
 #include "libfsm/internal.h" /* XXX */
+#include "libfsm/print/ir.h" /* XXX */
 
 #include "lx/ast.h"
 #include "lx/print.h"
 
 /* XXX: abstraction */
 int
-fsm_print_cfrag(FILE *f, const struct fsm *fsm,
+fsm_print_cfrag(FILE *f, const struct ir *ir, const struct fsm_options *opt,
```
silentbicycle: Once it's working with the IR, what is `struct fsm_options *opt` still needed for? Could that be stored within the IR instead?

src/libfsm/print/ir.h
katef: No, just the former. I was originally thinking of adding a mask of which strategies to allow, but I tried that and didn't like it much, and currently I think it'd make more sense to have options like "always make a table" set in `struct fsm_options` instead.

src/libfsm/print/ir.h
katef: These will always be bytes - for UTF-8, the ranges here describe the values for a single byte within a UTF-8 sequence. Whereas for the AST, i think we'll have unicode codepoints there.

src/libfsm/print/ir.h
katef: They're not public, no.

src/libfsm/print/ir.h
silentbicycle: Oh, so this is after the code points are chunked into bytes. Right.

src/lx/print/c.c
katef: For various rendering options, like "always hex". I could duplicate those to an equivalent `struct ir_options` containing just the relevant subset perhaps - I actually tried splitting the .c files here such that they don't include any of the FSM structs at all. But I decided there wasn't any benefit, since this is all internal anyway.

PR #95, opened 2018-07-13
Parallelize AST transformations in lx

dhobsd:


src/lx/Makefile
```diff
@@ -17,6 +17,8 @@ CFLAGS.${src} += -D LX_HEADER='"lexer.h"'
 DFLAGS.${src} += -D LX_HEADER='"lexer.h"'
 .endfor
 
+LFLAGS += -lpthread
```
katef: `LFLAGS.lx` for just lx

katef:
From discussion on IRC: Parsing time still dominates for relatively simple regexps:
```
; ./lx-pth -x example -x comments -QC 1 < ~/da-cz.lx > /dev/null
  364.28s user 0.50s system 99% cpu 6:06.29 total
; ./lx-pth -x example -x comments -QC 6 < ~/da-cz.lx > /dev/null
  379.94s user 0.73s system 100% cpu 6:18.23 total
```
But I suspect this has more benefit for complex regexps, where `fsm_determinise()` can get alarmingly expensive. We ought to be able to show that by fuzzing out .lx syntax with varying complexities.

PR #96, opened 2018-07-14
Port fuzz/gengraph.rb to awk, remove Ruby dependency.

silentbicycle:
This adds an awk script, `fuzz/gengraph`, which replaces the previous ruby script.

Note that this is not a 100% compatible port:

- The PRNG is different, so the results won't be identical.

- Rather than drawing a random count from a shuffled list to pick end
  nodes, his generates end node IDs and adds them to the random set.
  If any IDs are duplicated, it will end up with fewer end nodes than
  it would otherwise.

- It doesn't print the seed, num_nodes, etc. variables twice. That looks
  like a bug in the ruby script.

- The logic for generating edges is slightly different -- the ruby
  script generated a label for a unique (src,dst) pair, the awk script
  generates unique (src,dst,label) triples. In other words, the ruby
  script never generated multiple edges with distinct labels between any
  pair of nodes, but it seems like that's something that should be
  accounted for in testing/benchmarking.

- Rather than sorting the edge set, iterate through it with a double
  for loop in ascending order, because awk lacks a built-in sort.


katef:
For example:
```
; ./fuzz/gengraph $RANDOM 5 2 3 xyz 0.5 \
    | fsm -capl dot \
    | dot -Tpng -o /tmp/x.png /tmp/x.dot
```
x

PR #99, opened 2018-07-16
re: Fix missing error handling.

silentbicycle:
- Add an error case (`##`) for `list-of-alt` in the native and
  pcre dialects. The pcre dialect didn't have the same error because
  the case that explicitly allows empty alts (e.g. `(xyz|)`) was
  covering it.

- Add a test (tests/native/in17.re) which checks for the expected
  error message (and, implicitly, that it doesn't fail an assert).

- Add a case in `DIALECT_PARSE` that checks for a parse that
  returned success but has NULL for the head AST expression -- this
  indicates that there was an error, but the error handling is
  missing/incorrect. When built with asserts, it will assert(0),
  so fuzzers can help find any other missing cases. Otherwise,
  treat it like a parsing error, rather than failing an assert
  in re_ast_analysis.

- Add NULL checks in the semantic actions for all nodes that
  create/return an AST node.


src/libre/dialect/native/parser.sid
```diff
@@ -310,6 +310,8 @@
 			ALT;
 			r = list-of-alts();
 			node = <ast-expr-alt>(a, r);
+		##
+			node = <ast-expr-empty>();
```
katef: I think this should be `||` (and I suppose for the other dialects too), because it's not an error - it's a legal empty alt. For `##` I'd expect to print an error message and bail out.

silentbicycle: I figured that might not be the case, because the top of the grammar says:

    TODO: explain choices. e.g. i'm not permitting () or ||. only the root expression is allowed to be empty.

whereas PCRE explicitly does handle it.

silentbicycle: If the `atom` production's first part (sub/char class/literal/any) gets an `##` error case with `e = <ast-expr-empty>();` instead, then it parses, but produces an empty NFA. Is that preferable?

Also, `(*)` is the minimal case -- the problem is how `atom` should handle failing to match the first non-nullable part, before the count.

katef: Originally I'd decided (somewhat arbitrarily) to disallow empty alts for the native dialect, even if other dialects permit them. So I think `##` is the right thing to do. The comment at the top is talking about what's permitted for a successful parse, and this is not successful; an empty alt is a syntax error. So I added `<err-expected-alts>` here, since this is under `##`, and that's the right place to catch syntax errors.

PR #102, opened 2018-08-16
Add start and end anchors (`^` and `$`) to PCRE dialect

silentbicycle:
- Make CONCAT and ALT nodes into regular linked lists -- the right node is either another or EMPTY, ending the list.

- Unroll CONCAT and ALT into new CONCAT_N and ALT_N array nodes, because random access makes the anchoring analysis clearer.

- Keep CONCAT and ALT in the printer for now -- those can be removed entirely once CONCAT and ALT collect into arrays directly in the parser, but that's probably well out of scope for this PR.

- Add an ANCHOR type, which can be either START or END.

- Update the generated parser code, related to the above.

- Comment out the `tests/lxpos` makefile inclusion -- it seems to have been broken in the merge. This should get fixed soon, but I investigated a bit and didn't see any clear root cause.

- Address several other minor merge issues, add printing code for new node types, etc.

- Add several analysis passes, before compiling to an FSM -- these annotate nodes with various flags that influence state linking (whether states should link to the before/after nodes passed from the parent, or the global start/end or start/end-self-loop nodes), and flag subtrees or whole regexes that are inherently unsatisfiable.

- Add about 50 tests, under `tests/pcre-anchor/`.

- Update other PCRE tests so they are explicitly anchored, and update the expected results. Remove trailing newlines from all the tests (it was inconsistent).


sfstewman:
This is seriously exciting!  Thanks for all the hard work to get here.

src/libre/re.c
```diff
@@ -93,25 +94,35 @@ re_parse(enum re_dialect dialect, int (*getc)(void *opaque), void *opaque,
...
-		if (err != NULL) {
-			err->e = RE_EBADDIALECT;
-		}
+		if (err != NULL) { err->e = RE_EBADDIALECT; }
 		return NULL;
 	}
 
+	/* Needs explicit ^ $ anchors */
+	if (!m->implicitly_anchored) { flags |= RE_UNANCHORED; }
+
```
katef: :+1:

src/libre/re_analysis.c
```diff
@@ -5,78 +5,645 @@
...
+			}
+			return dst;
+		}
+	} else {
+		assert(0);
+		return NULL;
+	}
+}
+
+/* TODO: do this directly in the parser. */
```
katef: Would that get rid of the linked-list AST node type?

src/libre/re_ast.c
```diff
@@ -49,20 +77,34 @@ free_iter(struct ast_expr *n)
...
 	struct ast_expr *res = calloc(1, sizeof(*res));
 	if (res == NULL) { return res; }
 	res->t = AST_EXPR_EMPTY;
-
+	res->flags = RE_AST_FLAG_NULLABLE;
 	return res;
 }
 
+/* Note: this returns a single-instance node, which
+ * other functions should not modify. */
```
katef: Why not `const`?

src/libre/re_ast.h
```diff
@@ -31,13 +31,17 @@
...
+	AST_EXPR_ALT_N,
 	AST_EXPR_LITERAL,
 	AST_EXPR_ANY,
 	AST_EXPR_REPEATED,
 	AST_EXPR_CHAR_CLASS,
 	AST_EXPR_GROUP,
-	AST_EXPR_FLAGS
+	AST_EXPR_FLAGS,
+	AST_EXPR_ANCHOR,
+	AST_EXPR_TOMBSTONE
```
katef: I think the proper term for this is a "trap state", but I might be wrong.

src/libre/re_ast.h
```diff
@@ -90,34 +131,52 @@ struct ast_expr {
...
 			enum re_flags neg;
-			/* Previous flags, saved here and restored
-			 * when done compiling the flags node's subtree.
-			 * Since `struct ast_expr` contains a union,
-			 * this space is already allocated. */
-			enum re_flags saved;
 		} flags;
+		struct {
+			enum re_ast_anchor_type t;
+		} anchor;
```
katef: I'm surprised the anchors are independent nodes, rather than a bitmap attached to every node. Did you reject that idea already, or is it worth exploring?

src/libre/re_comp.c
```diff
@@ -24,6 +57,77 @@ comp_iter_repeated(struct comp_env *env,
...
+comp_iter(struct comp_env *env,
+    struct fsm_state *x, struct fsm_state *y,
+    struct ast_expr *n);
+
+static int
+can_skip_concat_state_and_epsilon(const struct ast_expr *l,
+    const struct ast_expr *r);
+
+static struct fsm *
+new_blank(const struct fsm_options *opt);
```
katef: Is this the same as `fsm_empty()`?

include/re/re.h
```diff
@@ -34,6 +34,7 @@ enum re_flags {
 	RE_REVERSE = 1 << 3,
 	RE_SINGLE  = 1 << 4,
 	RE_ZONE    = 1 << 5,
+	RE_UNANCHORED = 1 << 6,
```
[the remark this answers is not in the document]
silentbicycle: That shouldn't be a deep change, I think the main place it's checked (in re_comp, deciding anchoring) also has a flag indicating whether it's checking for the start or end edge.

src/libre/re_analysis.c
silentbicycle: Yes, and the linked-list ALT type as well.

src/libre/re_ast.c
silentbicycle: Because the const-ness of that particular node wouldn't mix with the non-const-ness of all the other node handling functions. Instead, the node free function skips it, as a special case, and the others just shouldn't modify. There could be more assertions, but it might be hard to express within the type system.

src/libre/re_ast.h
silentbicycle: I think it's "trap state" in automata literature, and "tombstone" in persistent data structure literature. I'm thinking about it more as the latter, because it's flagging a location in a structure whose previous value has been explicitly removed already and could be re-used, but doesn't force a global restructuring / rebalancing -- like a deleted entry in a hash table (that can be refilled, but is a distinct state from empty when searching), or a removed entry in a persistent tree or CRDT.

src/libre/re_ast.h
silentbicycle: I spent at least a week and a half on this specific issue.

My first several attempts converted them to flags on adjacent nodes, but that leads to some incredibly tricky cases when there are floating anchors, like `/(^)abc/`, or redundant but satisfiable anchors, like `/^^^abc/`.

Keeping them as independent nodes meant that they could behave like EMPTY nodes (x ---epsilon---> y), except the start or end is the global start or end, and then the epsilon closure algorithm during determinization already handles repeated satisfiable anchors correctly. Eventually, I figured that I was duplicating analysis it was doing later anyway.

It still needs to scan for unsatisfiable anchors (such as `/^a^bc/`, where it has non-nullable input before a non-nullable start anchor, and the corresponding end anchor case `/^ab$c$/`), because otherwise it ends up with false positives. Without explicit pruning, those would falsely match "bc" and "ab" respectively.

src/libre/re_comp.c
silentbicycle: `fsm_empty` returns whether the fsm is empty (it's a predicate), this returns a new FSM with a start state.

src/libre/re_analysis.c
```diff
@@ -5,78 +5,645 @@
...
+			/* Before recursing, save and restore whether
+			 * this is followed by anything which always consumes input.
+			 *
+			 * TODO: instead of doing a backward sweep for each node,
+			 * do a reverse falsification step as a separate pass? */
+			bak_consuming_after = env->followed_by_consuming;
+
+			/* If we don't already know whether this is being followed
+			 * by something that always consumes input, check */
+			if (!env->followed_by_consuming) {
```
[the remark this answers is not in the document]
silentbicycle: The tricky part here is that it can't just save whether there's an end anchor and reject the regex if it's followed by other nodes which must consume input -- the end anchor may be in an alt, and cases like `^abc(d|$)e$` should just prune the unsatisfiable anchor and still match otherwise.

Currently it's scanning more than it needs to (it could probably do it once and cache), but it seems to need a 'followed by consuming nodes' flag passed into the subtree.

src/libre/print/tree.c
```diff
@@ -40,48 +40,98 @@ fprintf_count(FILE *f, unsigned count)
...
 	case AST_EXPR_CONCAT:
 		fprintf(f, "CONCAT %p: {\n", (void *)n);
 		pp_iter(f, opt, indent + 1*IND, n->u.concat.l);
 		INDENT(f, indent);
+
 		fprintf(f, ", (%p)\n", (void *)n);
+
 		pp_iter(f, opt, indent + 1*IND, n->u.concat.r);
 		INDENT(f, indent);
 		fprintf(f, "} (%p)\n", (void *)n);
```
katef: Missing `break;` here I think (and below)

katef:
How the anchor nodes look:
image

src/libre/print/tree.c
silentbicycle: Good catch. Fixed in 6b56c40.

I deleted these, then added them back later (when I was debugging flattening the linked lists into arrays), but wound up not needing to check them, so I guess I didn't notice them missing the break.

Once the parser builds arrays directly, these cases should just be removed.

silentbicycle:
Fixed a bug caused by a now-unnecessary special case in a038c6a, added a regression test.

PR #104, opened 2018-08-25
Provide anchors for the native dialect

katef:
Here I invert `RE_UNANCHORED` so that a caller has the option of _not_ setting it (which is the opposite of what I'd originally thought we'd want). Then lx does set the `RE_ANCHORED` flag for its patterns in .lx files, and there's no way to un-set it for those.


src/libre/re.c
```diff
@@ -72,12 +72,13 @@ re_flags(const char *s, enum re_flags *f)
 		}
 
 		switch (*p) {
-		case 'i': *f |= RE_ICASE;   break;
-		case 'g': *f |= RE_TEXT;    break;
-		case 'm': *f |= RE_MULTI;   break;
-		case 'r': *f |= RE_REVERSE; break;
-		case 's': *f |= RE_SINGLE;  break;
-		case 'z': *f |= RE_ZONE;    break;
+		case 'a': *f |= RE_ANCHORED; break; /* PCRE's /x/A flag means /^x/ only */
```
silentbicycle: So, `/x/A` would still match "xyz"?

katef: Yep

PR #105, opened 2018-08-27
Test housekeeping

katef:
Just two things here:
* Removed trailing newlines from `.re` tests. These would be taken as part of the pattern under test, since these are binary files;
* Rewrote generated tests by hand. I'd like to keep the expected output as legible as possible, for sake of humans considering what's supposed to happen, when the behaviour changes.


tests/sql/out6.fsm
```diff
@@ -1,11 +1,12 @@
...
-4  ->  1 "a"; # e.g. "a"
-4  ->  2 "e"; # e.g. "e"
+
+0 -> 1 'a';
+1 -> 2 'b';
+0 -> 2;
+2 -> 0;
+
+2 -> 3 'e';
+3 -> 3 'e';
```
silentbicycle: Shouldn't it match the `f` in state 3, and then have a final state 4 that also epsilons back to 2?

The test NFA looks like it'd match `(ab)*e+`.

PR #114, opened 2019-02-20
C11 preprocessing tokenizer

jameysharp:
This is a larger sample input for `lx`, which may also be useful in its own right. I think it makes sense to ship it with libfsm, and I'm happy to license it under the same 2-clause BSD license that's given in `LICENCE`.

### DIFF
--- a/examples/lx/c11-pp.lx
+++ b/examples/lx/c11-pp.lx
@@ -1,0 +1,59 @@
+#
+# Copyright 2019 Jamey Sharp
+#
+# See LICENCE for the full copyright terms.
+#
+
+# Phase 3 of the C11 standard's translation phases covers decomposing a source
+# file into a sequence of preprocessing tokens. I believe this language
+# specification is a strict implementation of the standard.
+
+hexadecimal_digit = /[0-9a-f]/i;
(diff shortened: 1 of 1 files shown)


katef:
It's a wonderful example, thank you. I'll probably come back and add equivalent lexers for C89 and C99 too, when I can (unless you want to get there first). It'd be neat if they can produce the same tokens and share a superset for the header.

examples/lx/c11-pp.lx
```diff
@@ -0,0 +1,59 @@
...
+'?' | ':' | ';' | '...' |
+'=' | '*=' | '/=' | '%=' | '+=' | '-=' | '<<=' | '>>=' | '&=' | '^=' | '|=' |
+',' | '#' | '##' |
+'<:' | ':>' | '<%' | '%>' | '%:' | '%:%:' -> $punctuator;
+
+'/*' -> $block_comment_start .. '*/' -> $block_comment_end {
+	/./ -> $chr;
+}
+
+'//' -> $line_comment_start .. /\n+/ -> $newline {
```
katef: I want to see if I can borrow this trick for https://github.com/katef/kgt/issues/11

PR #117, opened 2019-02-22
Script to compile a demo lexer from a .lx spec

jameysharp:
I've been finding this very useful in playing with `lx`, so here it is in hopes it will be useful to others too.


katef:
Thanks!

PR #118, opened 2019-02-24
Simple `lx -l json` printer

jameysharp:
I don't know what I'm doing but with enough copy-paste from other parts of the source tree I made a thing!

### DIFF
--- a/src/lx/print/json.c
+++ b/src/lx/print/json.c
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2019 Jamey Sharp
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+#include <assert.h>
+#include <stdio.h>
+#include <errno.h>
+
+#include <adt/set.h>
+
+#include <fsm/fsm.h>
+#include <fsm/pred.h>
+
+#include "libfsm/internal.h" /* XXX */
+
+#include "lx/ast.h"
+#include "lx/print.h"
+
+/* TODO: centralise with libfsm */
+static unsigned int
+indexof(const struct fsm *fsm, const struct fsm_state *state)
+{
+	struct fsm_state *s;
+	unsigned int i;
+
+	assert(fsm != NULL);
+	assert(state != NULL);
+
+	for (s = fsm->sl, i = 0; s != NULL; s = s->next, i++) {
+		if (s == state) {
+			return i;
+		}
+	}
+
+	assert(!"unreached");
+	return 0;
+}
+
+static unsigned int
+zindexof(const struct ast *ast, const struct ast_zone *zone)
+{
+	struct ast_zone *z;
+	unsigned int i;
+
+	assert(ast != NULL);
+	assert(zone != NULL);
+
+	for (z = ast->zl, i = 0; z != NULL; z = z->next, i++) {
+		if (z == zone) {
+			return i;
+		}
+	}
+
+	assert(!"unreached");
+	return 0;
+}
+
+static int
+print_zone(FILE *f, const struct ast *ast, const struct ast_zone *z)
+{
+	struct fsm_state *s, *st;
+	int ret;
+
+	assert(f != NULL);
+	assert(z != NULL);
+	assert(z->fsm != NULL);
+	assert(fsm_all(z->fsm, fsm_isdfa));
+	assert(ast != NULL);
+
+	fprintf(f, "    {\n");
+	fprintf(f, "      \"initial_state\": %u,\n", indexof(z->fsm, z->fsm->start));
+	fprintf(f, "      \"states\": [\n");
+
+	for (s = z->fsm->sl; s != NULL; s = s->next) {
+		struct fsm_edge *e;
+		struct set_iter it;
+
+		fprintf(f, "        {\n");
+
+		e = set_first(s->edges, &it);
+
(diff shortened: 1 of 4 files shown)


src/lx/print/json.c
```diff
@@ -0,0 +1,163 @@
...
+	for (z = ast->zl; z != NULL; z = z->next) {
+		if (!fsm_all(z->fsm, fsm_isdfa)) {
+			errno = EINVAL;
+			return;
+		}
+	}
+
+	fprintf(f, "{\n");
+	fprintf(f, "  \"prefix\": { \"api\": \"%s\", \"tok\": \"%s\", \"lx\": \"%s\" },\n",
+		prefix.api, prefix.tok, prefix.lx);
```
katef: I feels a bit strange to output the prefixes here - these are to affect the generated code, not to be part of it. But I know you're intending to use this json to go and generate code, so it does make sense to include them here.

src/lx/print/json.c
```diff
@@ -0,0 +1,163 @@
...
+			assert(m != NULL);
+
+			if (m->token != NULL) {
+				fprintf(f, "          \"token\": \"$%s\",\n", m->token->s);
+			}
+			if (m->to != NULL) {
+				fprintf(f, "          \"next_zone\": %u,\n", zindexof(ast, m->to));
+			}
+
+			fprintf(f, "          \"accepts\": true%s\n", e != NULL ? "," : "");
```
katef: Terminology: libfsm probably should have its `.isend` field renamed to "accept" at some point. I'd like the naming to be consistent.

Likewise for your "initial_zone" field, where that's called either the starting zone (when talking about the graph), or the global zone (when talking about scope).

katef: And the same for "initial_state"

src/lx/print/json.c
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		if (e != NULL) {
+			int first = 1;
+			fprintf(f, "          \"transitions\": [");
+			for (; e != NULL; e = set_next(&it)) {
+				struct set_iter jt;
+				for (st = set_first(e->sl, &jt); st != NULL; st = set_next(&jt)) {
+					fprintf(f, "%s\n            { \"symbol\": %u, \"target\": %u }",
```
katef: I'm not sure about these numeric symbols. Do you think a single-letter string would be more fitting?

src/lx/print/json.c
katef: And for "target", which libfsm currently calls "dst" or "to".

src/lx/print/json.c
```diff
@@ -0,0 +1,163 @@
...
+		fprintf(f, "        {\n");
+
+		e = set_first(s->edges, &it);
+
+		if (fsm_isend(z->fsm, s)) {
+			const struct ast_mapping *m = s->opaque;
+			assert(m != NULL);
+
+			if (m->token != NULL) {
+				fprintf(f, "          \"token\": \"$%s\",\n", m->token->s);
```
katef: These strings should be escaped, per src/print/json.c's `json_escputc()`

src/lx/print/json.c
```diff
@@ -0,0 +1,163 @@
...
+
+			fprintf(f, "          \"accepts\": true%s\n", e != NULL ? "," : "");
+		}
+
+		if (e != NULL) {
+			int first = 1;
+			fprintf(f, "          \"transitions\": [");
+			for (; e != NULL; e = set_next(&it)) {
+				struct set_iter jt;
+				for (st = set_first(e->sl, &jt); st != NULL; st = set_next(&jt)) {
```
katef: libfsm has an IR for code generation which provides a few optimisations for generated DFA. I'm not sure if you'd want to use it here. It provides different "strategies" for each state, where e.g. symbols are presented grouped as `x..y` ranges, or e.g. if all symbols transition to the same state. If you're going on to generate code from the json, I think it probably would make sense to make use of that.

To use that IR, I'd first add json output to libfsm (cut & pasted from the current C), which just walks through that IR and outputs what it's given. The IR was introduced here: https://github.com/katef/libfsm/pull/94

Then `re -pl json xyz` would give json for a single regexp, too.

jameysharp:
Thanks for reviewing! Your feedback makes sense. I'm working on something else right now so I'm not sure when I'd get back to this, but I probably will eventually... :sweat_smile: For the moment, though:

I didn't realize this, but I guess `libfsm` and `re` can already output JSON&mdash;and apparently in two styles, with or without IR optimizations, at that? You have a comment in `libfsm/print/irjson.c` saying "TODO: leaf callback for json output", which I assume is the main thing that's missing for `lx` to be able to use it as well. Is there more to it than that?

katef:
So they do! Replace it with a better schema if you prefer.

The leaf callback is how lx renders things attached by its `void *opaque` - you can see that used for he generated C (to undertake zone changes) and for re's dot output (to print regexp indicies). It's a bit of an ugly arrangement, but it allows libfsm to print the graph, and callers to be responsible for rendering their own private concepts. I do think you'd want the same for json.

No worries about time! We're all busy with things. These APIs won't change, and I think this PR can wait as long as it needs.

katef:
See #198 for new json for libfsm - this doesn't include the lx parts.

PR #119, opened 2019-03-08
Introduce Allocation Function wrappers

Caslyn:
This PR introduces custom free/malloc/realloc function wrappers to be used for memory management and tracking in libfsm. The reason for this PR change is to allow user-provided allocation functions to define custom malloc/realloc/free directives.

The way this PR defines custom allocation wrapper functions is by introducing `f_free()`, `f_malloc()` and `f_realloc()` to use in the place of `free()`, `malloc()`, and `realloc()` respectively. The `f_*` functions take a `struct fsm *` along with the necessary data to perform the allocation: `fsm_options` is inspected (via the `struct fsm *`) and if it contains a user-defined allocation function, that function will be used, otherwise the default (libc) allocator is used. 

The following decisions formed this approach:

- Minimize the amount of consideration callers of the `f_` functions need in order to use them (ie minimize the amount of changes or data accesses around using these functions)
- Make the functions look familiar to the standard libc functions so that they can be easier to use.
- Future developers do not need to worry about setting default allocation functions when creating a `struct fsm_options` or `struct fsm`: the `f_` function will automatically take care of using the default libc functions if a user-defined allocator function is not available.
- Localize the custom user-defined allocation functions to `struct fsm *`: for example, if we want to move where/how we store function pointers within `struct fsm`, the necessary changes to do so can be confined to changing the `f_` function definition.

The changes are contained within src/libfsm - but if they need to be extended to other areas of the repository, let me know.

### DIFF
--- a/include/fsm/options.h
+++ b/include/fsm/options.h
@@ -82,7 +82,7 @@ struct fsm_options {
 		struct fsm *, struct fsm_state *);
 
 	/* closure for custom allocation functions */
-	struct fsm_allocator *allocator;
+	const struct fsm_allocator *allocator;
 };
 
 #endif
--- a/include/fsm/options.h
+++ b/include/fsm/options.h
@@ -82,7 +82,7 @@ struct fsm_options {
 		struct fsm *, struct fsm_state *);
 
 	/* closure for custom allocation functions */
-	const struct fsm_allocator *allocator;
+	const struct fsm_allocator allocator;
 };
 
 #endif
--- a/include/fsm/options.h
+++ b/include/fsm/options.h
@@ -18,6 +18,13 @@ enum fsm_io {
 	FSM_IO_PAIR
 };
 
+struct fsm_allocator {
+	void (*free)(void *opaque, void *p);
+	void *(*malloc)(void *opaque, size_t sz);
+	void *(*realloc)(void *opaque, void *p, size_t sz);
+	void *opaque;
+};
+
 struct fsm_options {
 	/* boolean: true indicates to go to extra lengths in order to produce
 	 * neater-looking FSMs for certian operations. This usually means finding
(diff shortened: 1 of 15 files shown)


include/fsm/options.h
```diff
@@ -73,6 +80,9 @@ struct fsm_options {
 	/* TODO: explain */
 	void (*carryopaque)(const struct fsm_state **, size_t,
 		struct fsm *, struct fsm_state *);
+
+	/* closure for custom allocation functions */
+	struct fsm_allocator *allocator;
```
katef: `const` maybe

src/libfsm/fsm.c
```diff
@@ -19,7 +19,57 @@
...
 	switch (0) { case 0: case (pred):; }
 
-static void
+void
+f_free(const struct fsm *fsm, void *p)
+{
+	const struct fsm_allocator *a;
+
+	assert(fsm != NULL);
+	assert(fsm->opt != NULL);
```
katef: And `fsm->opt->allocator != NULL`

include/fsm/options.h
Caslyn: 👍 2c08ea7

src/libfsm/fsm.c
Caslyn: The following lines cover the NULL scenario for `fsm->opt->allocator` - it applies the libc free if applicable.

sfstewman:
Wow.  This is great.  As @katef reminded me, I asked her for something like this at least a year ago.

katef:
> The changes are contained within src/libfsm

Not included: The ADT facilities, the CLI tools, libre. Of those, lx is the only other thing to make use of the ADT allocations. I think that's fine for our purposes. In practice this means some allocations will still come from the regular malloc and friends - but we always free those on error for library interfaces.

If we ever need to have all library allocations go through these callbacks, I think libre and the ADT allocations would cover it. I don't see any reason to do that here, and I think it'd be awkward to share that between the libraries.

Issue #103, opened 2018-08-24
portability: `clock_gettime` needs `-lrt` on Linux


silentbicycle:
`src/fsm/main.c:378: undefined reference to `clock_gettime'`

The makefile for `fsm` (the CLI util, not the library) should check `uname` and add `-lrt` to `LFLAGS.fsm` on Linux.

katef:
Which libc is this?

silentbicycle:
The default libc on Void, probably glibc. I've never tried to change it.

katef:
Curious. I don't see the same on debian:
```

; ldd gh/libfsm/build/bin/fsm                            
        linux-vdso.so.1 (0x00007ffefabd3000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdbc05a8000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fdbc07c5000)
; /lib/x86_64-linux-gnu/libc.so.6                        
GNU C Library (Debian GLIBC 2.28-8) stable release version 2.28.
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
Compiled by GNU CC version 8.3.0.
libc ABIs: UNIQUE IFUNC ABSOLUTE
For bug reporting instructions, please see:
<http://www.debian.org/Bugs/>.
; objdump -T /lib/x86_64-linux-gnu/libc.so.6 | grep clock_gettime
0000000000107360 g    DF .text  000000000000006d  GLIBC_PRIVATE __clock_gettime
0000000000107360  w   DF .text  000000000000006d  GLIBC_2.17  clock_gettime
```

silentbicycle:
According to the Linux man page for `clock_gettime`: 

    Link with -lrt (only for glibc versions before 2.17).

and glibc 2.17 is from 2013 or so. I've run into broken builds due to this within the last 5 or so years, probably because of systems that were still running older versions of glibc. It was something I encountered a couple times in CI, oddly enough.

PR #121, opened 2019-03-11
Link with `-lrt` for glibc < 2.17

katef:
For #103.

The logic here is probably overkill. I found linking with -lrt was harmless on my glibc 2.28 system, so I think this could probably be simply:
```
.if ${SYSTEM} == Linux
.if ${LDD_NAME} == glibc || ${LDD_NAME} == eglibc
LFLAGS.fsm += -lrt
.endif
.endif
```

But regardless the libc name still needs to be found, assuming some libcs don't come with a librt.


tbeebe:
Tested on glibc 2.12 successfully.

PR #125, opened 2019-03-20
Changes libfsm to use specific set types instead of `set` adt.

sfstewman:
Adds specific set types (state_set, edge_set, etc.) to libfsm.  This makes it easier to see which specific sets hold what values and allows the implementation of specific sets to vary.

This stacks on #124


sfstewman:
Rebased onto master after #124 merge.

Issue #120, opened 2019-03-09
lx IR failing on iscomplete predicate


katef:
Something definitely screwy here.
```
; echo '/[^b]/ -> $t;' | ~/gh/libfsm/build/bin/lx -l c

...

static enum lx_token
z0(struct lx *lx)
{
        int c;

        enum {
                S0, S1, NONE
        } state;

        assert(lx != NULL);

        if (lx->clear != NULL) {
                lx->clear(lx->buf_opaque);
        }

        state = NONE;

        lx->start = lx->end;

        while (c = lx_getc(lx), c != EOF) {
                if (state == NONE) {
                        state = S0;
                }

lx: src/libfsm/print/ir.c:422: make_state: Assertion `fsm_iscomplete(fsm, state)' failed.
```

jameysharp:
Once I started building with `DEBUG=1` I ran into this bug on lots of inputs. The simplest test I found that triggers this assert was `re -pl ir '^[^a]'`. Some cases which don't trigger the assert:

- If it isn't anchored at start-of string (`re -pl ir '[^a]'`)
- If the character class excludes more than one symbol (`re -pl ir '^[^ab]'`)
- And, as the source location suggests, if the output format doesn't use the IR (`re -pl json '^[^a]'`)

katef:
Currently thinking the `n == UCHAR_MAX + 1` in `fsm_iscomplete()` is incorrect, introduced here: https://github.com/katef/libfsm/commit/a4ee1690efba48738aae606aafe0ef7d509a0d4d

katef:
I think `n == UCHAR_MAX + 1` is correct, because `n` is the number of symbols, not the maximum symbol value. Confirmed that tests/pred/iscomplete*.fsm look correct by eye.

katef:
It's a fenceposting error introduced here: https://github.com/katef/libfsm/commit/b20bfef89e8e058ae6cb457a0ca8f722cd217e38#diff-019e5a5726da2e96b38dd52892399225R406
And a couple of other places where `UCHAR_MAX` is mixed with the number of elements in a set.

PR #128, opened 2019-03-23
Fenceposting error

katef:
This PR addresses the assertion reported in #120, reached by an off-by-one due to mixing the maximum value of the last element in a set (`UCHAR_MAX`) with the number of elements in a set (`UCHAR_MAX + 1`).


src/libfsm/print/ir.c
```diff
@@ -81,7 +81,7 @@ static struct ir_group *
 make_groups(const struct fsm *fsm, const struct fsm_state *state, const struct fsm_state *mode,
 	size_t *u)
 {
-	struct range ranges[UCHAR_MAX]; /* worst case */
+	struct range ranges[UCHAR_MAX + 1]; /* worst case */
```
sfstewman: Given that you've updated this in a few places, it seems like an enum constant might be safer here.  eg: 
```enum { MAX_RANGES = UCHAR_MAX + 1 };```

katef: I might do that sometime, but perhaps after I'm clearer about what the alphabet (Sigma) is for FSMs.

PR #132, opened 2019-03-26
Alternative error handling for `fsm_print_c`

Caslyn:
This PR returns `-1` instead of asserting or returning `EOF` in error scenarios.  

The reasons for these changes are: 
- Make error handling more flexible for the caller of the fsm function; for example, handling a `-1` return value instead of dying on an `assert` 
- Making `-1` a more explicit error value instead of `EOF`; reducing the need for the context to handle or understand an `EOF` value.
- Eliminate the need to include libraries for `EOF`/`assert` in the generated code

Without these changes, a context that wants to handle `NULL` (or empty string) input differently would need to maintain that handling separately (and have it executed before the libfsm-generated code is called); additionally it would be confined to using `.fragment = true` if it wants this error handling in the same function as the state machine.

### DIFF
--- a/src/libfsm/print/c.c
+++ b/src/libfsm/print/c.c
@@ -269,7 +269,7 @@ endstates(FILE *f, const struct fsm_options *opt, const struct ir *ir)
 
 	/* no end states */
 	if (!ir_hasend(ir)) {
-		printf("\treturn -1; /* error */\n");
+		printf("\treturn -1; /* unexpected EOT */\n");
 		return;
 	}
 
@@ -289,7 +289,7 @@ endstates(FILE *f, const struct fsm_options *opt, const struct ir *ir)
 		}
 		fprintf(f, "\n");
 	}
-	fprintf(f, "\tdefault: return -1; /* error */\n");
+	fprintf(f, "\tdefault: return -1; /* unexpected EOT */\n");
 	fprintf(f, "\t}\n");
 }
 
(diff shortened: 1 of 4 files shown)


src/libfsm/print/c.c
```diff
@@ -423,7 +421,9 @@ fsm_print_c(FILE *f, const struct fsm *fsm)
 		fprintf(f, "{\n");
 		fprintf(f, "\tint c;\n");
 		fprintf(f, "\n");
-		fprintf(f, "\tassert(fsm_getc != NULL);\n");
+		fprintf(f, "\tif (fsm_getc == NULL) {\n");
```
silentbicycle: Is fsm_getc being NULL here potentially a recoverable error?

src/libfsm/print/c.c
```diff
@@ -432,7 +432,9 @@ fsm_print_c(FILE *f, const struct fsm *fsm)
 		fprintf(f, "{\n");
 		fprintf(f, "\tconst char *p;\n");
 		fprintf(f, "\n");
-		fprintf(f, "\tassert(s != NULL);\n");
+		fprintf(f, "\tif (s == NULL) {\n");
+		fprintf(f, "\t\treturn -1;\n");
```
silentbicycle: Is this recoverable? If so, should it be distinct from the -1 `EOF` result?

src/libfsm/print/c.c
Caslyn: Undefined behavior it is! dc52d7c

src/libfsm/print/c.c
Caslyn: Undefined behavior it is! dc52d7c

PR #134, opened 2019-03-27
Collect unreferenced states after re_comp; add queue ADT.

silentbicycle:
This adds a garbage collection pass in `re_comp`, for states that end up being unreachable.

Also:
- add a fixed-capacity FIFO queue ADT, with tests
- make the contract for `equiv` in `struct fsm_state` more explicit, with a `tmp` union

This still needs direct calls to malloc/free/calloc switched over to the new memory management callback.


src/adt/queue.c
```diff
@@ -0,0 +1,81 @@
...
+#include <string.h>
+#include <assert.h>
+
+struct queue {
+	/* Read and write offsets. When rd == wr, the queue is empty. rd
+	 * will always be <= wr, and wr <= capacity. */
+	size_t rd;
+	size_t wr;
+	size_t capacity;
+	void *q[1 /* capacity */];
```
sfstewman: This would be a great place for a C99 flexible array member.  Maybe @katef will make a C99 exception?

src/libfsm/internal.h
```diff
@@ -43,14 +43,20 @@ struct fsm_edge {
...
-	 * meaningful within one particular transformation only */
-	struct fsm_state *equiv;
+	/* these are only valid within one particular transformation.
+	 * expected to be NULL at start and set back to NULL after. */
+	union {
+		/* temporary relation between one FSM and another */
+		struct fsm_state *equiv;
+		/* tracks which states have been visited in walk2 */
+		struct fsm_state *visited;
+	} tmp;
```
sfstewman: I'm glad to see this turned into a union; my prototype epsilon closure code has been abusing casts to stash other struct pointers in the `equiv` field, and this will let me clean that up fairly easily.

src/libfsm/collect.c
```diff
@@ -0,0 +1,108 @@
...
+	struct queue *q = queue_new(state_count);
+	int res = 0;
+	if (q == NULL) { return 1; }
+
+	if (!queue_push(q, fsm->start)) {
+		goto cleanup;
+	}
+	fsm->start->reachable = 1;
+	
+	for (;;) {
```
sfstewman: Maybe worth generalizing this loop into a general BFS walk?

src/adt/queue.c
katef: Yes, fine. We can have C99 per file :)

src/libfsm/collect.c
```diff
@@ -0,0 +1,108 @@
...
+{
+	const unsigned state_count = fsm_countstates(fsm);
+	struct queue *q = queue_new(state_count);
+	int res = 0;
+	if (q == NULL) { return 1; }
+
+	if (!queue_push(q, fsm->start)) {
+		goto cleanup;
+	}
+	fsm->start->reachable = 1;
```
katef: I'll move this to the predicate bitmap, when I'm closer to making a PR for that.

src/libre/re_comp.c
```diff
@@ -108,6 +109,8 @@ re_comp_ast(struct ast_re *ast,
 
 	if (!comp_iter(&env, x, y, ast->expr)) { goto error; }
```
katef: Would you want to remove states with no outgoing edges here, too?

The AST->NFA conversion sometimes leaves some of those, but I think they're only ever from the unused alternates for anchors. So it might be simple enough to keep track of whether they're used during the AST->NFA and know to remove them there instead.

src/libfsm/collect.c
silentbicycle: Kate and I talked about doing that for a couple other functions as well, but probably in a different PR.

src/libre/re_comp.c
silentbicycle: But then if those leave something else hanging, it will take another pass to remove those, ... and at that point we might be better off using reference counting and a zero-count table instead of mark and sweep.

katef: Hm that's interesting.

About the dead states: My comment here was specifically about the ones introduced from the AST's unused anchor/not-anchor states. For these, I _think_ the AST->NFA transformation has access to everything it needs to know in order to remove those without leaving anything else hanging. I'm picturing that as just keeping a boolean, saying which of the anchor/not-anchor states got used, but that might be over-simplified and miss-handle some situation I missed.

About reference counting in general: That'd be good for replacing `fsm_trim()`, which is expected to handle FSM passed in by a user.

katef: In other words, is there any way removing them _can_ leave something else hanging?

silentbicycle: For the dead-end start/end states? I don't believe so, I'm just trying to avoid making too much of a special case of them.

src/libfsm/collect.c
sfstewman: That makes sense.

Incidentally, I'm curious why you went with BFS here instead of a recursive DFS.  The BFS would be useful for Dijkstra's shortest path and similar algs, but for mark-and-sweep, you just want to visit reachable nodes, and a DFS should handle that.

src/libre/re_comp.c
katef: Ok, I don't mind either way. I do think this is a viable option, and should be cheap. I don't know what it means for complexity of code in that area though.

src/libfsm/collect.c
silentbicycle: Whether it's BFS or DFS is pretty arbitrary to me, my main concern was that it didn't overrun the stack when marking a huge graph recursively.

src/libfsm/walk2.c
```diff
@@ -250,7 +250,8 @@ fsm_walk2_tuple_new(struct fsm_walk2_data *data,
 	if (comb == NULL) {
 		return NULL;
 	}
-	comb->equiv = NULL;
+
+	assert(comb->tmp.visited == NULL);
```
katef: Why would this be true? As a side effect of setting another field in the union?

katef: Oh because `fsm_state_clear_tmp()` memsets the union to `0x0`. Okay. I'll merge this, but I'd prefer to come back and do that per field in the union instead (and the same for `memset()` elsewhere).

PR #140, opened 2019-04-09
Fixes the tail after fsm_collect_unreachable_states

sfstewman:


src/libfsm/collect.c
```diff
@@ -102,6 +103,14 @@ sweep_states(struct fsm *fsm)
 		}
 		s = next;
 	}
+
+	/* update the tail in case it was collected */
+	for (tail = &fsm->sl; *tail != NULL; tail = &(*tail)->next) {
+		continue;
+	}
+
+	fsm->tail = tail;
```
silentbicycle: Doesn't the `if (s == *fsm->tail) { *fsm->tail = prev; }` on line 96 take care of this?

While I haven't thoroughly checked, I suspect these overlap -- though your full pass update seems more obviously correct to me.

silentbicycle: I'm deep into other stuff right now, but could switch branches and look at this more closely later today. Walking `fsm->sl` and setting the tail in one pass seems like a better approach, though.

silentbicycle: In that case, wouldn't it suffice to remove the line updating `*fsm->tail` in the loop and add: ```
	if (fsm->sl != NULL) {
		assert(prev != NULL);
	}
	*fsm->tail = prev;
``` at the end? Because we already _were_ walking `fsm->sl`.

Comment (sfstewman) [reply src/libfsm/collect.c]: I think you're right!  But you need to update `fsm->tail`.  `fsm->tail` can point to either `fsm->sl` or the `next` member of a state.  So it's probably sufficient to do:
```
fsm->tail = (fsm->sl != NULL) ? &prev->next : &fsm->sl;
```

silentbicycle: I posted an alternate PR, #144, with a small change that makes the use-after-free go away, and also makes sense to me why.

I'm not set on merging that PR instead, but if we stick with this one, the code needs to be updated.

Thoughts?

sfstewman: PR #144 looks good to me, and I prefer it to this PR.  #144 avoids an extra (and slow) iteration through the linked list.

It's not clear to me that either address the case where the first state in `fsm->sl` is deleted.  I may write a few unit tests for that to make myself feel better.

sfstewman:
Closing, as #144 provides a better fix.

silentbicycle:
#144 should leave `tail` set to `&fsm->sl` in the case where it's empty.

sfstewman:
I agree.  My concern is different: if `fsm->sl` is swept, I don't see where `fsm->sl` will be updated in this case.

Consider someone using the libfsm API to:
1) Add several states.
2) Add edges, none of which point to the first state.
3) Set the start to be something other than the first state.
4) Runs `fsm_collect_unreachable_states`

This seems a legit use of the API, if a bit wasteful/odd.  I think `fsm->sl` will still point to a freed state after step 4.

PR #143, opened 2019-04-15
New, simpler and hopefully faster approach to building the dfa graph

sfstewman:
Eliminates `listnonepsilonstates` and `allstatesreachableby`, and instead builds the outgoing edges for each state of the dfa in a single pass.  This has a few advantages:

1. Reduces temporary allocations.  No `trans` structs are created or destroyed, reducing temporary churn.
2. Eliminates most `fsm_hasedge_literal` calls
3. Reduces `state_set_add` calls, hopefully avoiding some situations where building the state set becomes an expensive insertion sort.

NB: Commit d6f1fa3 is a temporary fix to the `fsm_trim` bug until @silentbicycle can put in a more permanent fix.


katef:
I also removed the bookkeeping for the (now unpopulated) trans set:
```diff
index bdf7f041..56c220f2 100644
--- a/src/libfsm/determinise.c
+++ b/src/libfsm/determinise.c
@@ -56,22 +56,8 @@ struct mapping {
 
 struct fsm_determinise_cache {
    struct mapping_set *mappings;
-   struct trans_set *trans;
 };
 
-static void
-clear_trans(const struct fsm *fsm, struct trans_set *trans)
-{
-   struct trans_iter it;
-   struct trans *t;
-
-   for (t = trans_set_first(trans, &it); t != NULL; t = trans_set_next(&it)) {
-       f_free(fsm, t);
-   }
-
-   trans_set_clear(trans);
-}
-
 static void
 clear_mappings(const struct fsm *fsm, struct mapping_set *mappings)
 {
@@ -86,20 +72,6 @@ clear_mappings(const struct fsm *fsm, struct mapping_set *mappings)
    mapping_set_clear(mappings);
 }
 
-static int
-cmp_trans(const void *a, const void *b)
-{
-   const struct trans *ta, *tb;
-
-   assert(a != NULL);
-   assert(b != NULL);
-
-   ta = a;
-   tb = b;
-
-   return (ta->c > tb->c) - (ta->c < tb->c);
-}
-
 static int
 cmp_mapping(const void *a, const void *b)
 {
@@ -474,7 +446,6 @@ determinise(struct fsm *nfa,
    struct mapping *curr;
    struct mapping_set *mappings;
    struct mapping_iter it;
-   struct trans_set *trans;
    struct fsm *dfa;
 
    struct mapping_iter_save sv;
@@ -507,17 +478,6 @@ determinise(struct fsm *nfa,
    }
    mappings = dcache->mappings;
 
-   if (dcache->trans == NULL) {
-       dcache->trans = trans_set_create(cmp_trans);
-       if (dcache->trans == NULL) {
-           fsm_free(dfa);
-           mapping_set_free(mappings);
-           return NULL;
-       }
-   }
-   trans = dcache->trans;
-   assert(trans != NULL);
-
    /*
     * The epsilon closure of the NFA's start state is the DFA's start state.
     * This is not yet "done"; it starts off the loop below.
@@ -617,13 +577,6 @@ fsm_determinise_cache(struct fsm *fsm,
            f_free(fsm, *dcache);
            return 0;
        }
-
-       (*dcache)->trans    = trans_set_create(cmp_trans);
-       if ((*dcache)->trans == NULL) {
-           mapping_set_free((*dcache)->mappings);
-           f_free(fsm, *dcache);
-           return 0;
-       }
    }
 
    dfa = determinise(fsm, *dcache);
@@ -646,16 +599,11 @@ fsm_determinise_freecache(struct fsm *fsm, struct fsm_determinise_cache *dcache)
    }
 
    clear_mappings(fsm, dcache->mappings);
-   clear_trans(fsm, dcache->trans);
 
    if (dcache->mappings != NULL) {
        mapping_set_free(dcache->mappings);
    }
 
-   if (dcache->trans != NULL) {
-       trans_set_free(dcache->trans);
-   }
-
    f_free(fsm, dcache);
 }
```
and then src/adt/transset.c can go, too.

src/adt/set.c
```diff
@@ -84,6 +84,87 @@ set_create(int (*cmp)(const void *a, const void *b))
...
+		free(s);
+		return NULL;
+	}
+
+	s->i = s->n = set->i;
+	memcpy(s->a, set->a, s->n * sizeof s->a[0]);
+
+	return s;
+}
+
```
katef: :+1:

src/libfsm/determinise.c
```diff
@@ -284,109 +284,171 @@ nextnotdone(struct mapping_set *mappings, struct mapping_iter_save *sv)
...
+
+					assert(outedges[sym].states != NULL);
+					assert(outedges[sym].len > 0);
+
+					/* add states from first edge */
+					for (i=0; i < outedges[sym].len; i++) {
+						if (!hashset_add(outsets[sym], outedges[sym].states[i])) {
+							goto finish;
+						}
+					}
```
katef: Lovely.

src/libfsm/determinise.c
```diff
@@ -284,109 +284,171 @@ nextnotdone(struct mapping_set *mappings, struct mapping_iter_save *sv)
...
 {
 	struct fsm_state *s;
 	struct state_iter it;
+	int sym, sym_min, sym_max;
+	int ret = 0;
 
-	assert(set != NULL);
-	assert(trans != NULL);
+	struct state_array outedges[FSM_SIGMA_COUNT];
+	struct hashset *outsets[FSM_SIGMA_COUNT];
```
katef: If we included the symbol in the hash function, could we use the same hashset for all edges here? Would that be worse?

src/libfsm/determinise.c
```diff
@@ -284,109 +284,171 @@ nextnotdone(struct mapping_set *mappings, struct mapping_iter_save *sv)
...
 
-/*
- * Return a list of all states reachable from set via the given transition.
- */
-static int
-allstatesreachableby(struct state_set *set, char c, struct state_set *sl)
-{
-	struct fsm_state *s;
-	struct state_iter it;
+	for (sym = sym_min; sym <= sym_max; sym++) {
```
katef: I'd `for (sym = 0; sym < UCHAR_MAX; sym++)` here rather than `<=`, for ease of grepping against the existing code elsewhere.

src/libfsm/internal.c
```diff
@@ -0,0 +1,100 @@
...
+	if (n > 0) {
+		states = state_set_array(src);
+		memcpy(dst->states, states, n * sizeof states[0]);
+	}
+
+	dst->len = n;
+
+	return dst;
+}
+
```
katef: Could we have this under /src/adt/something.c please? (Yes, I know it's not an ADT)

src/libfsm/internal.h
```diff
@@ -90,5 +90,23 @@ fsm_state_clear_tmp(struct fsm_state *state);
...
+
+struct state_array *
+state_array_add(struct state_array *arr, struct fsm_state *st);
+
+struct state_array *
+state_array_copy(struct state_array *dst, const struct state_array *src);
+
+struct state_array *
+state_array_copy_set(struct state_array *dst, const struct state_set *src);
+
```
katef: Likewise to include/adt/something.h

src/libfsm/determinise.c
sfstewman: I'm not sure.  That's worth benchmarking.

src/libfsm/determinise.c
sfstewman: That would add a lot of unnecessary iteration.  I kept track of the min/max sym out based on the observation that many states have a small number of edges that are close to each other.

src/libfsm/internal.c
sfstewman: Sure.

src/libfsm/internal.h
sfstewman: Sure.

src/libfsm/determinise.c
katef: For the record, we'd misunderstood each other here - i was talking about FSM_SIGMA_COUNT vs. a character value, not about the min/max endpoints.

katef:
Thank you as always!

PR #150, opened 2019-05-08
libfsm/print/c: Fix output destination.

silentbicycle:
This has been causing that return to get lost from generated code.


katef:
This is indeed the only case:
```
; grep 'printf(' src/libfsm/print/*.c | grep -v fprintf
src/libfsm/print/c.c:           printf("\treturn -1; /* unexpected EOT */\n");
```

PR #161, opened 2019-08-29
Fixes the return code handling of `fsm_trim()`

sfstewman:
Noticed this trying to get the tests working.  PR does a few things:

* removes `fsm_collect_unreachable_states` from `libfsm.syms`.  This makes macOS willing to link `libfsm.so`

* Changes the return handing of `fsm_trim()` to return 1 on success and -1 on error.  Previously it passed the return of `sweep_states()` along undisturbed, which returned -1 on error and the number of trimmed states (which could be zero) on success.

* Changed the return type of `sweep_states()` from `int` to `long` in case someone tries to trim a graph with more than `INT_MAX` states that can't be reached from start.

Tests now pass again, except for `pcre/in5.re` which is very reliable in its failure.

### DIFF
--- a/src/libfsm/libfsm.syms
+++ b/src/libfsm/libfsm.syms
@@ -93,6 +93,5 @@ fsm_print_cfrag # XXX: workaround for lx
 make_ir # XXX: workaround for lx
 free_ir # XXX: workaround for lx
 
-fsm_collect_unreachable_states
 fsm_clear_tmp
 fsm_state_clear_tmp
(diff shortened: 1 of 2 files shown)


src/libfsm/libfsm.syms
```diff
@@ -93,6 +93,5 @@ fsm_print_cfrag # XXX: workaround for lx
 make_ir # XXX: workaround for lx
 free_ir # XXX: workaround for lx
 
-fsm_collect_unreachable_states
```
katef: Thanks!

PR #163, opened 2019-09-21
Fixes bugs in set_add_bulk

sfstewman:
This fixes two bugs:
1) Used `qsort()` which passes a `void **` to the arguments of its compare function.  This isn’t compatible with `set->cmp`.  Instead, I've written a median-of-3 quicksort that is compatible with `set->cmp`
2) Removes duplicate elements after the sort.

### DIFF
--- a/src/adt/set.c
+++ b/src/adt/set.c
@@ -197,6 +197,90 @@ set_add(struct set *set, void *item)
 
 #define BULK_ADD_THRESH 1
 
+static
+void set_quicksort(void **items, size_t n, int (*cmp)(const void *a, const void *b))
+{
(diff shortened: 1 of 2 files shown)


sfstewman:
Rebased onto master after merging #164, added a test case that I initially forgot to add in the original PR.

katef:
https://github.com/katef/libfsm/pull/165 introduces the `item_t` typedef, giving a concrete type for each element in a set. I think this should fix (or rather, avoid) the `void **` bug this PR addresses.

katef:
This PR did two things; address the type, and de-duplicate entries. Closing in preference to #166, which re-implements the latter on top of the #165 refactoring.

PR #164, opened 2019-09-21
Fix for FLAGS node in re_ast_print_dot

sfstewman:
When running:
```re -p -r pcre -l ast '(?i).{0,2}(Tom|Sawyer|Huckleberry|Finn)$' | dot -Tpng > ast1.png```

dot gives the error:
```Error: bad label format {FLAGS|{+i|-```

The error *seems* to be caused by unclosed braces.  This PR adjusts the braces so:

1. FLAGS doesn’t have a brace around it
2. pos-flags and neg-flags are enclosed in separate {} pairs

### DIFF
--- a/src/libre/print/dot.c
+++ b/src/libre/print/dot.c
@@ -160,13 +160,12 @@ re_ast_print_dot(FILE *f, const struct fsm_options *opt,
 static void
 re_flags_print(FILE *f, enum re_flags fl)
 {
-	const char *sep = "";
-	if (fl & RE_ICASE) { fprintf(f, "%si", sep); sep = " "; }
-	if (fl & RE_TEXT) { fprintf(f, "%sg", sep); sep = " "; }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/libre/print/dot.c [hunk 1]
```diff
@@ -125,11 +125,11 @@ pp_iter(FILE *f, const struct fsm_options *opt,
...
 
 	case AST_EXPR_FLAGS:
-		fprintf(f, "\tn%p [ label = <{FLAGS|{+", (void *) n);
+		fprintf(f, "\tn%p [ label = <FLAGS|{+", (void *) n);
 		re_flags_print(f, n->u.flags.pos);
-		fprintf(f, "|-");
+		fprintf(f, "}|{-");
 		re_flags_print(f, n->u.flags.neg);
-		fprintf(f, "> ];\n");
+		fprintf(f, "}> ];\n");
```
katef: Thanks! I have no idea what was going on here.

While you're here, could you remove the `sep` character for between flags please? So the output is just like `+imgs` etc.

PR #166, opened 2019-10-23
Revised bugfixes for the `set_bulk_add`

sfstewman:
1. Removes duplicates after the sort.

2. Includes a test case.


katef:
Thank you!

PR #173, opened 2019-11-11
Adds support for {M,} qualifiers to PCRE dialect

sfstewman:


sfstewman:
Replaced by #179, which has the advantages of (a) being more complete, and (b) being rebased onto the current master.

PR #174, opened 2019-11-17
Construct pre-generated named character classes in situ, rather than by merging into the constructed FSM

katef:
Here the main change is to switch the `-l api` code generation to pass in an existing FSM (with x,y endpoints), rather than to merge and union from a separately constructed FSM.

To facilitate this, I've done quite a bit of refactoring on libre's AST. Previously there were special-purpose nodes for character classes (and the earlier PR https://github.com/katef/libfsm/pull/172 worked to normalize these to nary nodes).

This PR continues from that earlier work to remove the character class nodes entirely. To do that, I've introduced one new node for subtraction, and both SQL 2003's `[abc^xyz]` syntax and the more common `[^abc]` set inversion syntax are implemented in terms of this.

Prior to #172:
image

and as of this PR:
image


src/libre/ast_analysis.c
```diff
@@ -133,23 +135,23 @@ analysis_iter(struct analysis_env *env, struct ast_expr *n)
...
 		break;
 
 	case AST_EXPR_ANCHOR:
 		/* anchor flags will be handled on the second pass */
 		break;
 
+	case AST_EXPR_SUBTRACT:
+		/* XXX: not sure */
+		break;
+
```
katef: @silentbicycle ^ this bit here :)

src/libre/ast_analysis.c
```diff
@@ -492,23 +514,27 @@ assign_lasts(struct ast_expr *n)
...
 	case AST_EXPR_GROUP:
-		set_flags(n, AST_EXPR_FLAG_LAST);
+		set_flags(n, AST_FLAG_LAST);
 		assign_lasts(n->u.group.e);
 		break;
 
+	case AST_EXPR_SUBTRACT:
+		/* XXX: not sure */
+		break;
+
```
katef: @silentbicycle also this

src/libre/ast_analysis.c
silentbicycle: I think this is okay, because an empty character class with subtraction is rejected by the parser.

silentbicycle: Put another way, in this spot this should be handled the same way as LITERAL and NAMED, because it will only transitively contain those as leaf nodes. (Right?)

src/libre/ast_analysis.c
silentbicycle: It seems like this and the corresponding case on line 471 should set the FIRST/LAST flags and recurse to their contents, like with REPEATED/GROUP, but I'm struggling to come up with a test case that proves that it matters. It may be a non-issue because they can only ultimately contain non-nullable items? And input regexes like `(^a|[^[:digit:]]$)c$` seem to handle pruning unsatisfiable parts correctly.

PR #179, opened 2019-11-27
Assorted PCRE improvements - qualifiers on counts, and special cases for character class syntax

sfstewman:
- Adds support for `{M,}` qualifier
- Adds support for qualifiers-on-qualifiers (eg: `a{3,6}?`)
- Adds support for closing-bracket as the first letter of a group (eg: `[]abc]`, `[]-a]`, `[^]abc]`, and `[^]-a]`)
- Adds support for `-` as the first character of a group (eg: `[-abc]`, `[--0]`, `[^-abc]`, and `[^--0]`)

Notes:
1. This doesn't add support for `-` as the last character of a group.
2. This PR moves a lot of lexer logic into the parser, largely because there's no way to tell lx to handle the first characters of a group differently.  
3. Replaces #173


sfstewman:
Rebased onto master, resolving a bunch of accumulated conflicts.

Note that `tests/pcre/in5.re` is slightly broken because the error message has changed.

PR #181, opened 2019-11-29
Fixes how REPEATED nodes are translated to an NFA

sfstewman:
Came upon an error where determinising `/a{2,2}/` clearly gave the wrong graph.  The problem was in the NFA generated by the AST:
image

Interestingly, `/a{1,1}/` was correct:
image

And `/aa/` was correct:
image

This narrowed the issue down to the translation of AST_EXPR_REPEATED that called `fsm_state_duplicatesubgraphx`.  After some investigation, the problem turned out to be that when the subexpression of AST_EXPR_REPEATED is created, it can be linked to the global self-loop for unanchored regexps.  These links are then repeatedly copied by `fsm_state_duplicatesubgraphx`.

The subexpressions are linked to the global self-loop because they're identified as FIRST / LAST expressions in the `assign_firsts()` / `assign_lasts()` functions.  This PR attempts to fix the problem by not recursively assigning firsts/lasts for the subexpression of an AST_EXPR_REPEATED node.  This fixes the particular problem:
image

**However, I am concerned that this may break something else.**


sfstewman:
These commits have been incorporated into #182.  I'm leaving this open for now because it has a much better description of the bug than #182.

src/libre/ast_analysis.c
```diff
@@ -459,7 +459,16 @@ assign_firsts(struct ast_expr *n)
...
+		/* Don't recurse.
+		 *
+		 * XXX - Not sure that this is the correct way to handle this.
+		 *
+		 * Recursing causes errors in the NFA when the REPEATED node is
+		 * linked to the global self-loop (ie: it's either the first or
+		 * last node for an unanchored RE.  Specifically, when the
+		 * subexpression is compiled, the links to the global self-loop
+		 * are created, which the REPEATED node then copies.
+		 */
```
katef: @silentbicycle I'm merging this now, but I'd like your view here.

PR #182, opened 2019-12-04
Assorted fixes and improvements to libre

sfstewman:
Contains several fixes and one improvement.

Fixes:
1. Fixes NFA generation for `AST_EXPR_REPEATED` nodes with explicit counts (eg: `/a{2,3}/`).  See #181 for a full description.
2. Fixes a realloc() bug in `ast_compile.c`
3. Fixes NFA generation for `AST_EXPR_ALT` nodes with repeated subnodes, eg: `(b+|a)`
4. Attempts to fix the interaction between the global self-loop and `AST_EXPR_SUBTRACT`.

For 1, 3, and 4, I am somewhat concerned that the changes may have broken something else.  I've done some testing.  In addition to the bundled unit tests, I've been working on running `re` against the pcre test suite.  Many of the above fixes came from bugs that the pcre test suite identified.

Improvement:
1. Adds an additional way to duplicate the subgraph that's much faster than `fsm_state_duplicatesubgraphx()`.

This incorporates and replaces #181.  I'll leave that PR open for now, though, since it contain a better description of the that bug than this PR does.


src/libfsm/subgraph.c
```diff
@@ -182,3 +182,125 @@ fsm_state_duplicatesubgraphx(struct fsm *fsm, fsm_state_t state,
...
+}
+
+void
+fsm_subgraph_capture_stop(struct fsm *fsm, struct fsm_subgraph_capture *capture)
+{
+	assert(fsm != NULL);
+	assert(capture != NULL);
+
+	capture->end = fsm_countstates(fsm);
+}
```
katef: This is a neat idea, at least while we have suitable backing storage to keep it efficient. I like it.

sfstewman:
Split into separate PRs for bugfixes and improvements (#183 and #184, respectively) as requested by @katef

PR #183, opened 2019-12-06
Assorted fixes to libre for anchoring and repetition

sfstewman:
Splits out the bug fixes in #182 into a separate PR, as requested by @katef 

Contains several fixes:

Fixes:
1. Fixes NFA generation for `AST_EXPR_REPEATED` nodes with explicit counts (eg: `/a{2,3}/`).  See #181 for a full description.
2. Fixes a realloc() bug in `ast_compile.c`
3. Fixes NFA generation for `AST_EXPR_ALT` nodes with repeated subnodes, eg: `(b+|a)`
4. Attempts to fix the interaction between the global self-loop and `AST_EXPR_SUBTRACT`.

For 1, 3, and 4, I am somewhat concerned that the changes may have broken something else.  I've done some testing.  In addition to the bundled unit tests, I've been working on running `re` against the pcre test suite.  Many of the above fixes came from bugs that the pcre test suite identified.

This, along with #184, replaces #132.  It incorporates and replaces #181.  I'll leave #181 open for now, though, since it contain a better description of the that bug than this PR does.


sfstewman:
Note that 2df9244 has some overlap with #184.

src/libre/ast_analysis.c
```diff
@@ -173,6 +173,10 @@ always_consumes_input(const struct ast_expr *n, int thud)
 	case AST_EXPR_NAMED:
 		return 1;
 
+	case AST_EXPR_SUBTRACT:
+		/* FIXME: need to perform the subtraction to know */
+		return 1;
+
```
katef: This worries me a bit. Is there ever a correct way to find this here?

I don't see how to avoid subtraction in the AST given that at least one dialect has an operator for it.

sfstewman: The main edge case here is if the subtraction gives the empty set.

katef: Can we have a test which hits this case? I think that'd be either in the SQL dialect using `[x^x]` (to produce an empty set) or in the native dialect using `[^\x00-\xff]` to achieve the same.

katef: Indeed these both fail on master:
```
; bmake -r CC=gcc DEBUG=1 && ./build/bin/re -r sql -pl dot '[x^x]'
re: src/libre/re.c:169: re_comp: Assertion `err == NULL || err->e != RE_ESUCCESS' failed.
zsh: abort      ./build/bin/re -r sql -pl dot '[x^x]'
; bmake -r CC=gcc DEBUG=1 && ./build/bin/re -r native -pl dot '[^\x00-\xff]'
re: src/libre/re.c:169: re_comp: Assertion `err == NULL || err->e != RE_ESUCCESS' failed.
zsh: abort      ./build/bin/re -r native -pl dot '[^\x00-\xff]'
; 
```

katef: Okay, I addressed these assertions in #187, so we should be able to use `[^\x00-\xff]` or equivalent in the contexts which'd reach this `/* FIXME */`.

sfstewman:
Rebased onto master

src/libre/ast_compile.c
```diff
@@ -520,17 +520,32 @@ comp_iter_repeated(struct comp_env *env,
...
 	} else if (low == 0 && high == 1) {                    /* '?' */
-		RECURSE(x, y, n->e);
-		EPSILON(x, y);
+		NEWSTATE(na);
+		NEWSTATE(nz);
+		EPSILON(x,na);
+		EPSILON(nz,y);
+
+		RECURSE(na, nz, n->e);
+		EPSILON(na, nz);
```
katef: It's a shame we can't avoid the epsilons here. But okay.

PR #184, opened 2019-12-06
Improves duplicating subgraphs (speeding up AST_EXPR_REPEATED node -> NFA translation)

sfstewman:
This adds an additional way to duplicate the subgraph that's much faster than `fsm_state_duplicatesubgraphx()`.

This splits out the improvement from #182 into a separate PR, as requested by @katef.

Along with #183, this replaces #182.


sfstewman:
This has some overlap with 2df9244 in #183.

PR #192, opened 2019-12-25
Initial version of the dfavm

sfstewman:
dfavm is a bytecode based way to run DFA matching.  This commit provides:

1. Initial cut at the dfavm "compiler", dfavm bytecode, and dfavm vm to run the bytecode.
2. `retest`, a program to test dfavm matching of regular expressions.
3. Two utility programs to adapt the PCRE test suites to something `retest` can run

### DIFF
--- a/format_pcre_tests.sh
+++ b/format_pcre_tests.sh
@@ -1,0 +1,15 @@
+#! /bin/sh
+
+jq --raw-output \
+    '.[]                              | 
+     select(.parsed)                  | 
+     [ .pattern, 
+       (.matches|map("+" + .)|join("\n")), 
+       (.not_matches|map("-" + .)|join("\n")) 
+     ]                                | 
+     select(.[0] != null)             |
+     select(.[1] != "" or .[2] != "") | 
+     [ (.[]|select(. != "")) ]        |
+     join("\n") + "\n"                |
+     @text' $@
+
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -314,6 +314,7 @@ xopen(const char *s)
 	return f;
 }
 
+#if 0
 static unsigned int word_maxlen = 16;
 static unsigned int num_words = 8;
 static uint64_t seed = 0xdfa7231bc;
@@ -336,6 +337,7 @@ gen_words(FILE *f, const struct fsm *fsm)
 
 	fsm_generate_words_to_file(fsm, &params, &prng, num_words, f);
 }
+#endif /* 0 */
 
 
 int
@@ -394,9 +396,11 @@ main(int argc, char *argv[])
 			case 't': op = op_name(optarg);               break;
 			case 'G': op = op_name("glushkovise");        break;
 			case 'W':
-				print = gen_words;
-				num_words = strtoul(optarg, NULL, 10);
+				/* print = gen_words; */
+				/* num_words = strtoul(optarg, NULL, 10); */
 				/* XXX: error handling */
+				fprintf(stderr, "not yet implemented.\n");
+				exit(EXIT_FAILURE);
 				break;
 
 			case 'h':
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -27,6 +27,7 @@
 #include "libfsm/internal.h" /* XXX */
 
 #include "parser.h"
+#include "wordgen.h"
 
 #if defined(__APPLE__) && defined(__MACH__) && defined(MACOS_HAS_NO_CLOCK_GETITME)
 #include <mach/clock.h>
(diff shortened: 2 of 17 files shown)


sfstewman:
The dfavm compiler does some amount of basic block reordering, but further optimization will come in a later PR.

format_pcre_tests.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/sh
```
katef: You can also do like
```
#!/usr/bin/jq -srf
```

src/fsm/main.c
```diff
@@ -312,6 +314,32 @@ xopen(const char *s)
...
+	struct prng_state prng;
+
+	memset(&prng, 0, sizeof prng);
+	prng_seed(&prng, seed);
+
+	fsm_generate_words_to_file(fsm, &params, &prng, num_words, f);
+}
+#endif /* 0 */
+
+
```
katef: What's going on here? Is `fsm_generate_words_to_file()` just not present yet?

src/libfsm/dfavm.c
```diff
@@ -0,0 +1,1692 @@
...
+#include <string.h>
+
+#include <fsm/fsm.h>
+#include <fsm/vm.h>
+#include "internal.h"
+#include "print/ir.h"
+
+#define DEBUG_ENCODING     0
+#define DEBUG_VM_EXECUTION 0
+#define DEBUG_VM_OPCODES   0
```
katef: I'm not sure about the encoding and execution, but I do think `DEBUG_VM_OPCODES` should be exposed to both `re(1)` and `fsm(1)` as a `-l ...` output language option. (You can treat the `fsm_print` function pointer essentially as an enum for that)

src/libfsm/dfavm.c
```diff
@@ -0,0 +1,1692 @@
...
+	case VM_OP_BRANCH:
+		{
+			char dst;
+			switch (op->u.br.dest) {
+			case VM_DEST_NONE:  dst = 'Z'; break;
+			case VM_DEST_SHORT: dst = 'S'; break;
+			case VM_DEST_NEAR:  dst = 'N'; break;
+			case VM_DEST_FAR:   dst = 'F'; break;
+			default:            dst = '!'; break;
+			}
```
katef: I do agree that the near/far jumps are probably overkill, but not strongly :)

src/libfsm/dfavm.c
```diff
@@ -0,0 +1,1692 @@
...
+fsm_vm_match_buffer(const struct fsm_dfavm *vm, const char *buf, size_t n)
+{
+	struct vm_state st = state_init;
+	enum dfavm_state result;
+
+	vm_match(vm, &st, buf, n);
+	result = vm_end(vm, &st);
+
+	return result == VM_SUCCESS;
+}
```
katef: Maybe not on this initial PR, but it'd probably be sensible to split this into several files, for compilation, execution, etc.

src/retest/main.c
```diff
@@ -0,0 +1,793 @@
...
+	if (cpy == NULL) {
+		if (err != NULL) {
+			*err = -1;
+		}
+		return NULL;
+	}
+
+	memcpy(cpy, s, len+1);
+	return cpy;
+}
```
katef: There's an `xstrdup()` in adt/xalloc.c if you want it

format_pcre_tests.sh
sfstewman: Nice!

Semi-relatedly, I'm thinking of scrapping this script and either adapting the `extract_pcre_tests.py` script to generate `retest` format directly or writing a small C program to do this (removing the python dependency).

src/fsm/main.c
sfstewman: Oh, whoops!  That was part of my on-hold project to generate text directly from a DFA.  Must have slipped in there...

src/libfsm/dfavm.c
sfstewman: That makes sense.

I'm thinking we probably want to split things so there's a high-level "IR-ish" set of opcodes and a bytecode encoding.  The opcodes would be useful for optimization and would be the basis for any generated bytecode and assembly.

In this scheme, you'd want `fsm_print` functions for both the opcodes and bytecodes.

src/libfsm/dfavm.c
sfstewman: As soon as I have a performance test suite, I'm going to try both 16-bit and 32-bit fixed width bytecodes for the VM.

src/libfsm/dfavm.c
sfstewman: Yes!  This file is getting too large.

PR #209, opened 2020-06-29
Various IR opcode fixes

sfstewman:
1. Makes debug printing more pleasant
2. Removes `-L ir_pre` flag and changes `-L ir` to print IR opcodes (after optimizations)
3. Fixes bug in `eliminate_unnecessary_branches` that leaves `num_incoming` incremented after removing a branch.

### DIFF
--- a/src/libfsm/vm/ir.c
+++ b/src/libfsm/vm/ir.c
@@ -837,6 +837,9 @@ struct dfavm_op_ir **find_opchain_end(struct dfavm_op_ir **opp)
 }
 
 static void
+dump_states(FILE *f, struct dfavm_assembler_ir *a);
+
+static void
 eliminate_unnecessary_branches(struct dfavm_assembler_ir *a)
 {
 	int count;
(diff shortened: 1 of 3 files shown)


katef:
Lovely seeing only the labels which are used :)
```
; ./build/bin/re -k str -pl vmc 'fi+sh'

int
fsm_main(const char *s)
{
        const char *p;
        int c;

        p = s;

l0:
        if (c = *p++, c == '\0') return -1;
        if (c != 'f') goto l0;
l1: /* e.g. "f" */
        if (c = *p++, c == '\0') return -1;
        if (c == 'f') goto l1;
        if (c != 'i') goto l0;
l2: /* e.g. "fi" */
        if (c = *p++, c == '\0') return -1;
        if (c == 'f') goto l1;
        if (c == 'i') goto l2;
        if (c != 's') goto l0;
        if (c = *p++, c == '\0') return -1;
        if (c == 'f') goto l1;
        if (c != 'h') goto l0;
        return 0x1; /* "fi+sh" */
}
```

PR #210, opened 2020-06-29
Go backend

dgryski:
Open tasks:
    

- [x] generate gofmt'd output by default
- [x] figure out what the package name should be
- [x] probably export the matching function by default
- [x] solve "unused labels are a compile error" issue

### DIFF
--- a/src/libfsm/print/go.c
+++ b/src/libfsm/print/go.c
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2008-2020 Katherine Flavel
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <stdint.h>
+#include <errno.h>
+#include <stdio.h>
+
+#include <print/esc.h>
+
+#include <adt/set.h>
+
+#include <fsm/fsm.h>
+#include <fsm/pred.h>
+#include <fsm/walk.h>
+#include <fsm/print.h>
+#include <fsm/options.h>
+#include <fsm/vm.h>
+
+#include "libfsm/internal.h"
+
+#include "libfsm/vm/vm.h"
+
+#include "ir.h"
+
+static int
+leaf(FILE *f, const void *state_opaque, const void *leaf_opaque)
+{
+	assert(f != NULL);
+	assert(leaf_opaque == NULL);
+
+	(void) state_opaque;
+	(void) leaf_opaque;
+
+	/* XXX: this should be FSM_UNKNOWN or something non-EOF,
+	 * maybe user defined */
+	fprintf(f, "return TOK_UNKNOWN;");
+
+	return 0;
+}
+
+static const char *
+cmp_operator(int cmp)
+{
+	switch (cmp) {
+	case VM_CMP_LT: return "<";
+	case VM_CMP_LE: return "<=";
+	case VM_CMP_EQ: return "==";
+	case VM_CMP_GE: return ">=";
+	case VM_CMP_GT: return ">";
+	case VM_CMP_NE: return "!=";
+
+	case VM_CMP_ALWAYS:
+	default:
+		assert("unreached");
+		return NULL;
+	}
+}
+
+static void
+print_label(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt)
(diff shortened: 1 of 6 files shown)


katef:
Thank you!!

dgryski:
Pushed an example to https://github.com/dgryski/trifles/tree/master/fsmdemo

```
~/go/src/github.com/dgryski/trifles/fsmdemo $ cat main.go
package main

//go:generate sh -c "re -r pcre -k str -e fish  -pl go 'fi+sh' >fishfsm/fishfsm.go"

import (
	"fmt"

	"github.com/dgryski/trifles/fsmdemo/fishfsm"
)

func main() {
	net := []string{"fsh", "catfish", "dogo", "fiiish"}
	for _, caught := range net {
		if fishfsm.Match([]byte(caught)) != -1 {
			fmt.Println("caught a fish: ", caught)
		}
	}
}
~/go/src/github.com/dgryski/trifles/fsmdemo $ go generate
~/go/src/github.com/dgryski/trifles/fsmdemo $ cat fishfsm/fishfsm.go
package fishfsm

func Match(data []byte) int {
	var idx = ^uint(0)

l0:
	if idx++; idx >= uint(len(data)) {
		return -1
	}

	if data[idx] != 'f' {
		goto l0
	}

l1: // e.g. "f"
	if idx++; idx >= uint(len(data)) {
		return -1
	}

	if data[idx] == 'f' {
		goto l1
	}

	if data[idx] != 'i' {
		goto l0
	}

l2: // e.g. "fi"
	if idx++; idx >= uint(len(data)) {
		return -1
	}

	if data[idx] == 'f' {
		goto l1
	}

	if data[idx] == 'i' {
		goto l2
	}

	if data[idx] != 's' {
		goto l0
	}

	if idx++; idx >= uint(len(data)) {
		return -1
	}

	if data[idx] == 'f' {
		goto l1
	}

	if data[idx] != 'h' {
		goto l0
	}

	{
		return 4
	}

}
~/go/src/github.com/dgryski/trifles/fsmdemo $ go run main.go
caught a fish:  catfish
caught a fish:  fiiish
```

src/libfsm/print/go.c
```diff
@@ -0,0 +1,273 @@
...
+		assert("unreached");
+		return NULL;
+	}
+}
+
+static void
+print_label(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt)
+{
+	fprintf(f, "l%lu:", (unsigned long) op->index);
+
```
katef, line 68: It's worth a comment here to explain that c_escputc_str's is a subset of go's strings, so we can get away with using that

src/libfsm/print/go.c
```diff
@@ -0,0 +1,273 @@
...
+	}
+}
+
+static void
+print_cond(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt)
+{
+	if (op->cmp == VM_CMP_ALWAYS) {
+		return;
+	}
+
```
katef, line 82: (same for the escaping here)

src/libfsm/print/go.c
```diff
@@ -0,0 +1,273 @@
...
+		uint32_t l;
+
+		l = 0;
+
+		for (op = a.linked; op != NULL; op = op->next) {
+			if (op->num_incoming > 0) {
+				op->index = l++;
+			}
+		}
+	}
```
katef, line 169: I guess I should centralise this!

src/libfsm/print/go.c
```diff
@@ -0,0 +1,273 @@
...
+
+	fprintf(f, "func Match");
+
+	switch (fsm->opt->io) {
+	case FSM_IO_STR:
+		fprintf(f, "(data []byte) int {\n");
+		/* start idx at -1 unsigned so after first increment we're correct at index 0 */
+		fprintf(f, "\tvar idx = ^uint(0)\n");
+		fprintf(f, "\n");
+		break;
```
katef, line 260: I might move this under something other than `FSM_IO_STR` (since I'd expect that to emit go's `string` type). Generating `[]byte` is great, but I want to call it something else, and I want to come back and revisit how `FSM_IO_*` are named.

PR #211, opened 2020-06-29
Add invariant: `fsm_minimise` should only be called with DFAs.

silentbicycle:
This invariant needs to be in place before switching to the new minimization algorithm.

I feel a bit apprehensive about adding a big invariant like this after the fact, but Brzozowski's algorithm is calling `fsm_determinise` internally anyway, albeit after reversing.

### DIFF
--- a/include/fsm/fsm.h
+++ b/include/fsm/fsm.h
@@ -278,7 +278,7 @@ fsm_complete(struct fsm *fsm,
 	int (*predicate)(const struct fsm *, fsm_state_t));
 
 /*
- * Minimize an FSM to its canonical form.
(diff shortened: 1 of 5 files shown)


sfstewman:
To chime in, this is a very sensible invariant.

Minimization of NFAs is considered NP-hard, and the minimum is not always uniquely definable.  Most algorithms focus on DFAs, including Brzozowski, Hopcroft, and the various hashing algorithms (including Liu et al).

There's actually almost no work on minimizing NFAs.  I occasionally look, because reducing the state space, even if not to a minimum, could help speed up determinization.

PR #213, opened 2020-07-03
Add Go assembly output

dgryski:
Just setting this here.  Remaining items

- [x] add code for matching against `[]byte`
- [x] determine if I can make asmfmt happy without breakage ( https://github.com/klauspost/asmfmt )


dgryski:
Turns out https://github.com/katef/libfsm/pull/213/commits/c3cac014f386fd7bbb9128b03fd32893d0b9c38c is pretty ugly and I don't think I can make it 100% because of how some jumps are generated and how asmfmt wants them laid out.  I probably won't keep this commit.

dgryski:
Demo updated for asm machine: https://github.com/dgryski/trifles/commit/7ec52febaef4471443f4dd9f712d68755dac0639

PR #212, opened 2020-06-30
Basic amd64 assembly generation

sfstewman:
It needs some work, but the basic structure is there.

PR #214, opened 2020-07-04
Two small Go assembly fixes:

dgryski:


dgryski:
This silly bug was the result of only testing with tiny regexps with few states. I have a better test suite I'm working in that compares the output of the Go and Go assembly backends with the vmc backend for larger more complicated expressions.

sfstewman:
> This silly bug was the result of only testing with tiny regexps with few states. I have a better test suite I'm working in that compares the output of the Go and Go assembly backends with the vmc backend for larger more complicated expressions.

@dgryski  I'd highly recommend running through the test suite I linked in https://github.com/katef/libfsm/pull/212#issuecomment-653343576 in  with `retest` if you can.  While not particularly exhaustive, that test suite (derived from the pcre test suite) has caught numerous bugs during development of dfavm.

katef:
Is it possible to have `retest` run the go-generated assembly, without introducing a build-time dependency on external packages?

dgryski:
Unlikely. We'd need something to parse Go flavoured ASM and transform it into something that could be assembled and called from C (with a trampoline to emulate Go's calling convention).

Probably easier to either introduce the dependencies (or, and this is the approach I'm taking) is to have the testing framework live outside of the main repo.

sfstewman:
Forgive me; I'll be bold enough to offer my $.02:

- I think it's fair for `retest` to call out to the Go assembler for this.  For vmasm, `retest` calls out to `as` (or `${AS}`), and for `c` and `vmc`, `retest` calls out to `cc` (or `${CC}`).
- Given the simplicity of the data passed in/out and the minimal use of the stack, I think writing a trampoline would be okay.  
- To me, the hard part would be linking to generate a `.so` file that `dlopen()` can load. It's been a few years since I've written any amount of Go, so I'm not sure what the state of dynamic library creation is in.  If we can run the Go assembler and linker, and generate a `.so` file that `dlopen` can load, the trampoline approach is viable.
- Whether this approach is fragile or not is a different question; keeping tests under `go test` is certainly less fragile.

dgryski:
Go *should* be able to generate a `.so` file that works with `dlopen()`; It's the `c-shared` buildmode, and would require a bit more hackery to get working.  It's not just calling the Go assembler with the generated code; you have to pull in the rest of the Go runtime (even though it wouldn't be used).  Obviously this requires the full Go toolchain to be installed.  I'm leaning towards reimplementing a testing framework, possibly using Go's plugin architecture or something..

dgryski:
I've been playing with creating a Go shared library to call an assembly function.  I'll update this with a link to a working example when I'm done and we can figure out how to get it into `retest`.

dgryski:
Turns out it's not that hard.  Here is a `main.go` file for my demo repo:

```
package main

import "github.com/dgryski/trifles/fsmdemo/fishasm"

import "C"

//export fishasmMatch
func fishasmMatch(base *C.char, len C.int) C.int {
	return C.int(fishasm.Match(C.GoStringN(base, len)))
}

func main() {}
```

That gets built with `go build -buildmode=c-shared` and then this runner from the `dlsym(3)` man page:

```
#include <stdio.h>
#include <stdlib.h>
#include <dlfcn.h>

int
main(int argc, char **argv)
{
    void *handle;
    int (*match)(char *s, int len);
    char *error;

   handle = dlopen("../sotest", RTLD_LAZY); /* fix path */
    if (!handle) {
        fprintf(stderr, "%s\n", dlerror());
        exit(EXIT_FAILURE);
    }

   dlerror();    /* Clear any existing error */

   *(void **) (&match) = dlsym(handle, "fishasmMatch");

   if ((error = dlerror()) != NULL)  {
        fprintf(stderr, "%s\n", error);
        exit(EXIT_FAILURE);
    }

    printf("%d\n", match("catfishies", 10)); /* 4  */
    printf("%d\n", match("dogo", 4));        /* -1 */
    dlclose(handle);
    exit(EXIT_SUCCESS);
```

katef:
I'm going to merge this for now (for the fixes, thank you!), and then come back and do the dlopen work separately.

katef:
@mshahdadi44 thank you for the review :)

PR #222, opened 2020-07-20
Update `fsm_trim`: Trim dead-end states.

silentbicycle:
Several updates, mostly related to `fsm_trim`:

- Actually trim states that have no path to an end state. The old comment incorrectly implied trim already did this.

- Rather than incrementally deleting each trimmed state (with an expensive pass over the whole graph to update edges), remap/compact edges in one pass, then compact the live states. This uses the new `fsm_compact_states` function.

- Fix bugs in `edge_set_remove` and `edge_set_remove_state` -- these adjusted the offset while modifying the array, so they skipped the entry immediately after the deleted one. This could trigger assert failures for edges leading to removed states. These functions are now only used by `fsm_mergestates`.

- Add a missing prototype for `f_calloc`.

- Update some comments.

Note: There is one failing test case with this, `build/tests/subtract-cuc/res2:FAIL` -- this exposed a bug in `fsm_complete`, which should be handled outside this PR.


src/adt/edgeset.c
```diff
@@ -349,12 +349,15 @@ edge_set_remove(struct edge_set **setp, unsigned char symbol)
...
-	for (i = 0; i < set->i; i++) {
+	i = 0;
+	while (i < set->i) {
 		if (set->a[i].symbol == symbol) {
 			if (i < set->i) {
 				memmove(&set->a[i], &set->a[i + 1], (set->i - i - 1) * (sizeof *set->a));
 			}
 			set->i--;
+		} else {
+			i++;
```
katef, line 360: Amazing that despite all the tests, we still didn't hit this

silentbicycle: For what it's worth, the way I found this was using `gengraph` to generate several corpora of NFAs in ascending sizes, used `fsm -pd` to turn them into DFAs, and then running the new (work in progress) minimization code on them. Some `assert` failures started showing up, and with a corpus of several thousand more small DFAs I eventually got a few that  triggered the asserts, yet were small enough (<10 states or so) to understand and reduce by hand.

PR #227, opened 2020-07-29
Convert edge set ADT to a hash table.

silentbicycle:
This converts the `edge_set` ADT to a hash table, because `fsm_minimise` will be spending a lot of time searching for individual edges by symbol. The existing uses based on iteration would not benefit from switching over -- they still need to check every edge. The `edge_set_contains` integrity assertions should become cheaper, though.

Also:

- Remove the `alloc` pointer from every individual edge state. It was passed in to all code paths anyway, so storing an extra copy per state just wastes memory.

- Fix the IR range combining -- the previous implementation assumed the edge set's edges were sorted by symbol. This isn't always true, though it has been true often enough by construction that tests were depending on it. Storing in hash table order exposed it. Now the IR ranges collected are sorted before combining.

- Add an ordered iterator (ordered by symbol), and update the printing code to use it.

### DIFF
--- a/src/libfsm/print/ir.c
+++ b/src/libfsm/print/ir.c
@@ -77,13 +77,11 @@ make_groups(const struct fsm *fsm, fsm_state_t state,
 	assert(state < fsm->statecount);
 	assert(u != NULL);
 
-	/*
-	 * The first pass here populates ranges[] by collating consecutive symbols
-	 * which transition to the same FSM state. These ranges are constructed by
-	 * iterating over symbols, and so are populated in symbol order.
-	 * So several ranges may transition to the same state; these are grouped
-	 * in the second pass below.
-	 */
+	/* The first pass gathers ranges, and opportunistically combines
+	 * adjacent ranges with consecutive symbols which transition to
+	 * the same FSM state. Since the edges are stored unordered (in
+	 * a hash table), a follow-up pass sorts and then combines any
+	 * newly adjacent ranges to the same state. */
 
 	n = 0;          /* number of allocated ranges */
 	grp_ind = 0;    /* index of current working range */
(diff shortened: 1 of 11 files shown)


src/libfsm/print/ir.c
```diff
@@ -106,13 +104,29 @@ make_groups(const struct fsm *fsm, fsm_state_t state,
...
+			} else {
+				assert(dst < src);
+				dst++;
+				memcpy(&ranges[dst], &ranges[src],
+				    sizeof(ranges[dst]));
+			}
+		}
+		n = dst + 1;
+	}
+
```
katef, line 124: Would you want to move this up to an earlier stage, when we construct the IR?

If there are entries we can combine, everything would benefit (at the cost of the sort, of course. but that's fine). And I'd like the IR printing to represent exactly what the structs contain, to make debugging less confusing.

silentbicycle: If we can do so unambiguously, earlier would be better. Should it be part of this PR or separate?

katef:
I find this is indistinguishable for `fsm_determinise()` of 0..5000 words of the same lengths (these will spread out very early, because the prefixes don't overlap much):
image
(`main` at the top, this branch at the bottom)

katef:
And with random word lengths of 50..100 (i.e. with more initial prefix overlap), I do see a consistent improvement - but not by much, for this case (times here are in ms):
image

katef:
Pathological case for `fsm_determinise()`, before and after:
image

(and you can see `fsm_glushkovise` is so small it's irrelevant here)

silentbicycle:
I wouldn't expect this to directly benefit those yet -- I checked and didn't see cases where there was clearly iteration for search that could be replaced by a single `edge_set_find`, whereas that case will completely dominate minimization. I may spot more opportunities to take advantage of the hash table backing in determinisation later though.

katef:
With `SET_INITIAL` set to 4, no significant difference:
image

src/libfsm/print/ir.c
silentbicycle: I just switched this to ordered iteration. I decided to leave the range consolidation in; it may not longer be necessary, but I'm not sure.

PR #229, opened 2020-08-04
Fixes a few bugs in the new fsm_trim() code

sfstewman:
While trying to use the new `fsm_trim()`, I encountered a few problems:

1. `fsm_trim` is too efficient; it trims away the whole graph when there are no accepting states.  This breaks a number of tests.  For now, I've added a hack that avoids trimming the starting state.  This fixes the tests that rely on that behavior.

2. `fsm_trim` wasn't collecting all edges in the forward pass.  If the destination of an edge had already been visited, that edge would be omitted.  This can result in backward paths being omitted, which can cause states to be trimmed on the reverse pass.

One example is in `tests/minimise/in11.fsm`, whose graph is:
min-in11

Running

```./build/bin/fsm -t trim -p -l dot tests/minimise/in11.fsm | dot -Tpng``` 

on the `main` branch produces:
min-in11-t

This PR fixes the behavior, and avoids trimming the states.

### DIFF
--- a/src/libfsm/trim.c
+++ b/src/libfsm/trim.c
@@ -133,10 +133,13 @@ mark_states(struct fsm *fsm, enum fsm_trim_mode mode)
 					goto cleanup;
 				}
 				fsm->states[next].visited = 1;
+			}
 
-				if (edges == NULL) {
-					continue;
-				}
+			/* have to add all edges whether we've already visited
+			 * them or not to ensure that all possible backward
+			 * paths are present in the edge list.
+			 */
+			if (edges != NULL) {
 				if (!save_edge(fsm->opt->alloc,
 					&edge_count, &edge_ceil, &edges,
 					s_id, next)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/libfsm/trim.c [hunk 1]
```diff
@@ -343,6 +349,17 @@ fsm_trim(struct fsm *fsm, enum fsm_trim_mode mode)
...
+		/* - XXX -
+		 * this is a hack to make sure that we leave at least the start state.
+		 * otherwise a bunch of tests break...
+		 */
+		fsm_state_t start;
+		if (fsm_getstart(fsm, &start)) {
+			fsm->states[start].visited = 1;
+		}
+	}
+
```
katef, line 362: We talked this through a bit (and also with @silentbicycle, who found the same bug). My view is:

I would prefer that we have trim remove the state, and deal with whatever needs fixing up for the other operations. It seems like removing it is the correct thing for trim.

That gives:
> fsm_determinise: Invalid argument

...which we can fix by saying that an empty FSM is deterministic, and therefore acceptable for fsm_determinise to return, and therefore acceptable to be passed in. I would expect this applies to fsm_minimise and fsm_glushkovise (to be renamed), too.

When that's permitted, @sfstewman says some of the tests don't terminate. So we'll need to figure out what's going on there.

silentbicycle: I'm not sure how the `fsm_determinise: invalid argument` case should be handled. For the time being, I've told my property tests to consider it uninteresting and skip it.

silentbicycle:
I made pretty much the same fix in #232 but I didn't need that start state change to make tests pass.

katef:
I'm closing this because as discussed above, we're keeping the behaviour for `fsm_trim()` to potentially remove a start state. If this is a problem for determinise or anything else, then those can address their own issues with it. We're also deciding to allow `fsm_determinise` to accept and return an empty FSM, and therefore call that deterministic.

PR #232, opened 2020-08-07
Add property tests for edge_set and fsm_trim

silentbicycle:
This adds some new property tests, fixed bugs found by them, and also does some misc. cleanup:

- Update some existing property tests to reflect interface changes. These tests aren't part of the default build, so the broken build went unnoticed.
- Reformat the property test CLI's usage message.
- Fix a parameter for the LCG, used for shuffling input data.
- Make `fsm_state_remap_fun`s opaque argument `const`.
- Fix an unused argument warning.
- Remove a warning in the makefile; the property tests are `-std=c99`.
- Add conditionally controlled logging for `fsm_trim`.
- Fix incremental rebuilding of the `build/theft/theft` target.
- Add a property test for edge_set operations, add several regression tests for bugs found, fix them. 
- Add a property test for trimming, a regression test for its minimal failing case, and fix the bug.

I am still working on property testing `fsm_minimise`, but that work should probably go in a later PR to avoid excessively large merge conflicts. Some of the other existing property tests need further attention, but at the moment I'm focusing on testing that will support `fsm_minimise`. I will address those later. In particular, the `nfa_minimise...` test is broken because `fsm_minimise` rejects NFAs now, and the `adt_set` tests are pointing out that the set does not actually compare sets with elements reordered as equal. (Since nothing currently relies on that, we are probably just going to remove `set_cmp`.)

The fix for the bug in `fsm_trim` is the same as in PR #229, except it does not have the XXX start state hack; the tests all pass without it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ RE     ?= re
 BUILD  ?= build
 PREFIX ?= /usr/local
 
-.if make(fuzz)
+.if make(fuzz) || make(${BUILD}/theft/theft)
 PKG += libtheft
 .endif
 
@@ -73,7 +73,7 @@ SUBDIR += tests/hashset
 SUBDIR += tests/queue
 SUBDIR += tests/aho_corasick
 SUBDIR += tests
-.if make(fuzz)
+.if make(fuzz) || make(${BUILD}/theft/theft)
 SUBDIR += theft
 .endif
 SUBDIR += pc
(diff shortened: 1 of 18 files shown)


sfstewman:
As a note: the "need" for the start hack showed up when I added a trim step while developing/testing a version of hopcroft minimization.  I'm not surprised that it's not needed without that.

I was mostly being focused (or perhaps lazy) and wanted to finish what I was working on, so I didn't even try to fix it the proper way (as @katef subsequently requested).

katef:
#235 handles the absence of start state

PR #233, opened 2020-08-12
`fsm_minimise`: Add property test, switch to Moore's algorithm (with a couple optimizations)

silentbicycle:
The history for this is somewhat gnarly, there was some rebasing onto main to incorporate other changes (edge_set hash tables, other property tests) and there have been some false starts and intermediate commits squashed out.

- Add a property test for `fsm_minimise`: for any generated DFA specification, the number of states in the DFA produced by `fsm_minimise` should agree with the result from a simpler-but-inefficient algorithm, here used as a test oracle. This found inputs where the Brzozowski-based `fsm_minimise` actually added states -- see `dfa_minimise_regression_brz`.

- The liveness check previously used by the `fsm_trim` property test has been extracted so the `fsm_minimise` property test can use it for trimming the test data.

- Add `fsm_consolidate`, which takes a `struct fsm *` and a new state mapping, then produces a new `struct fsm *` with states renumbered and/or combined according to the mapping.

- Switch the `fsm_minimise` implementation from Brzozowski's algorithm to Moore's. This is substantially faster in typical cases, and the worst case (a single very deeply nested path to an end state) can be easily detected ahead of time.

- Add a couple small algorithmic improvements (summarized in `minimise_internal.h`: see `done_ec_offset` and `SMALL_EC_THRESHOLD`) that serve roughly the same purpose as the splitter queue management in Hopcroft's algorithm, but don't become significantly more expensive as k (the alphabet size) increases the way Hopcroft's will.

This is a draft, because:
- I'm not sure that `fsm_consolidate` is using `fsm_carryopaque_array` correctly, and I'm not sure how to set up a good standalone test for it.
- Many of the tests are currently failing, because the semantics around start states in empty DFAs have changed slightly -- `tests/native/in24.re` is representative. Still figuring this out.

### DIFF
--- a/src/libfsm/consolidate.c
+++ b/src/libfsm/consolidate.c
@@ -29,7 +29,7 @@ struct mapping_closure {
 };
 
 static fsm_state_t
-mapping_cb(fsm_state_t id, void *opaque)
+mapping_cb(fsm_state_t id, const void *opaque)
 {
 	const struct mapping_closure *closure = opaque;
 	assert(id < closure->count);
--- a/src/libfsm/consolidate.c
+++ b/src/libfsm/consolidate.c
@@ -23,6 +23,8 @@
 
 #include "internal.h"
 
+#define LOG_MAPPING 0
+
 struct mapping_closure {
 	size_t count;
 	const fsm_state_t *mapping;
@@ -44,6 +46,7 @@ fsm_consolidate(struct fsm *src,
 	fsm_state_t src_i;
 	uint64_t *seen = NULL;
 	struct mapping_closure closure;
+	size_t max_used = 0;
 
 	assert(src != NULL);
 	assert(src->opt != NULL);
@@ -53,10 +56,20 @@ fsm_consolidate(struct fsm *src,
 		goto cleanup;
 	}
 
-	if (!fsm_addstate_bulk(dst, mapping_count)) {
+	for (src_i = 0; src_i < mapping_count; src_i++) {
+#if LOG_MAPPING
+		fprintf(stderr, "consolidate_mapping[%u]: %u\n",
+		    src_i, mapping[src_i]);
+#endif
+		if (mapping[src_i] >= max_used) {
+			max_used = mapping[src_i];
+		}
+	}
+
+	if (!fsm_addstate_bulk(dst, max_used + 1)) {
 		goto cleanup;
 	}
-	assert(dst->statecount == mapping_count);
+	assert(dst->statecount == max_used + 1);
 
 	seen = f_calloc(src->opt->alloc,
 	    mapping_count/64 + 1, sizeof(seen[0]));
@@ -74,11 +87,6 @@ fsm_consolidate(struct fsm *src,
 
 	closure.count = mapping_count;
 	closure.mapping = mapping;
-
-	for (src_i = 0; src_i < mapping_count; src_i++) {
-		fprintf(stderr, "mapping[%u]: %u\n",
-		    src_i, mapping[src_i]);
-	}
 
 	for (src_i = 0; src_i < mapping_count; src_i++) {
 		const fsm_state_t dst_i = mapping[src_i];
--- a/src/libfsm/consolidate.c
+++ b/src/libfsm/consolidate.c
@@ -100,14 +100,14 @@ fsm_consolidate(struct fsm *src,
 				dst->opt->alloc, src->states[src_i].epsilons)) {
 				goto cleanup;
 			}
-			/* FIXME compact mapping! */
+			state_set_compact(&dst->states[dst_i].epsilons,
+			    mapping_cb, &closure);
 
 			if (!edge_set_copy(&dst->states[dst_i].edges,
 				dst->opt->alloc,
 				src->states[src_i].edges)) {
 				goto cleanup;
 			}
-			/* FIXME compact mapping! */
 			edge_set_compact(&dst->states[dst_i].edges,
 			    mapping_cb, &closure);
 
@@ -115,11 +115,36 @@ fsm_consolidate(struct fsm *src,
 				fsm_setend(dst, dst_i, 1);
 			}
 		} else {
-			assert(fsm_isend(dst, dst_i) == fsm_isend(src, src_i));
+			const int is_end = fsm_isend(dst, dst_i);
+			fsm_state_t prev_i;
+			fsm_state_t prev_src_id = src->statecount; /* NONE */
+			fsm_state_t src_set[2];
 
-			/* src->opt->carryopaque(src,  */
-			/* TODO carry opaque */
-			/* can we do a pairwise carry? */
+			if (!is_end) {
+				continue;
+			}
+			assert(fsm_isend(src, src_i));
+
+			/* Find the previous state that also maps to this
+			 * state, to carry the opaque from it. This could
+			 * become expensive when there are a very high number
+			 * of states, so it may be worth keeping a temporary
+			 * second mapping for just the end states. */
+			for (prev_i = 0; prev_i < mapping_count; prev_i++) {
+				if (mapping[prev_i] == dst_i) {
+					prev_src_id = prev_i;
+					break;
+				}
+			}
+			assert(prev_src_id < src->statecount);
+
+			src_set[0] = prev_src_id;
+			src_set[1] = src_i;
+
+			/* call carryopaque pairwise */
+			fsm_carryopaque_array(src,
+			    src_set, 2,
+			    dst, dst_i);
 		}
 	}
 
--- a/src/libfsm/consolidate.c
+++ b/src/libfsm/consolidate.c
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2020 Scott Vokes
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+/* fsm consolidate -> take an FSM and a mapping, produce a new FSM with combined states */
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <fsm/fsm.h>
+#include <fsm/pred.h>
+#include <fsm/walk.h>
+
+#include <adt/alloc.h>
+#include <adt/set.h>
+#include <adt/edgeset.h>
+#include <adt/stateset.h>
+#include <adt/hashset.h>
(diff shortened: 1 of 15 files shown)


silentbicycle:
Just making `fsm_determinise` return success and an empty FSM for an input NFA without a start state conceptually makes sense, and it also makes all the tests pass except for `tests/ir/in[57].re`, both of which seem to produce output which looks equivalent but has renumbered states.

silentbicycle:
Rebased onto main (with #235), added a commit which updates the expected IR test output. This should be ready for review now.

src/libfsm/minimise.c
```diff
@@ -7,49 +7,660 @@
...
+		}
+	}
+	assert(*label_count == count);
 }
 
+/* Build a mapping for a minimised version of the DFA, using Moore's
+ * algorithm (with a couple performance improvements).
+ *
+ * For a good description of the algorithm (albeit one that incorrectly
+ * attributes it to Hopcroft) see pgs. 55-59 of Cooper & Torczon's
```
katef, line 172: :joy:

src/libfsm/minimise.c
```diff
@@ -7,49 +7,660 @@
...
+		fprintf(stderr, "# --init[%lu]: ec %d, jump[]= %d\n",
+		    i, ec, env->jump[i]);
+#endif
+	}
+
+	/* The dead state is not a member of any EC. */
+	env->state_ecs[env->dead_state] = NO_ID;
+}
+
+#if EXPENSIVE_INTEGRITY_CHECKS
```
katef, line 392: Great idea, thanks

katef: Do you think it's worth keeping this separate to `-DNDEBUG`? We have debugging assertions that are already super expensive, and I don't expect a `make DEBUG=1` build to perform well at all.

dhobsd: This would be my vote if I may cast my possibly worthless two cents.

katef: Of course you may!! Always happy to have your input!

silentbicycle: On my own projects, I prefer to be more explicit and use `assert` for checks I expect to always run, and conditionally compile other more expensive per-loop invariants behind a `#define` like this. I very rarely use `-DNDEBUG`, because the performance gains in doing so have always been small (after wrapping asserts I know will be expensive), and it usually leads to unused variable warnings. Generally, I would wrap any assertion that would have a non-constant cost (such as sweeping through a collection to make sure something is present/removed) in an `#if` like this, but keep assertions on for things like bounds or NULL checks.

I left it this way so we could have this conversation during code review. The property tests aren't as meaningful with `-DNDEBUG`, though -- by design they're thrashing around in the state space and trying to provoke assert failures being used as internal consistency checks, which often catch problems quicker than tests on the results, so if we do those should be built separately.

dhobsd: This is an excellent counterpoint. At the same time, I feel like EXPENSIVE_INTEGRITY_CHECKS might be not the best name for this (in that it is perhaps too specific). If NDEBUG is too generic, perhaps there's a suitable compromise? Or even something super obvious like -DPROPERTY_TESTS

silentbicycle: I don't think of them as being specific to the property tests, though. They're integrity, internal consistency, or invariant checks that can affect the algorithmic order of growth, rather than constant-time checks whose overhead can (hopefully) mostly be eliminated by branch prediction. Maybe `EXPENSIVE_ASSERTS`?

sfstewman: My $.02, for what they're worth... 

afaik, libfsm doesn't have checks that would match the role of `EXPENSIVE_INTEGRITY_CHECKS`/`EXPENSIVE_ASSERTS` in any other place.  I'd rather keep the existing name, push for more expensive checks like these find their way into the code, and then try to generalize.  It's easier to find a better name when you have more than one concrete example.

silentbicycle: I had these two in mind when I wrote the above:
- https://github.com/katef/libfsm/blob/main/src/adt/edgeset.c#L237
- https://github.com/katef/libfsm/blob/main/src/adt/edgeset.c#L472

and there are others, such as:
- https://github.com/katef/libfsm/blob/main/src/libfsm/determinise.c#L88
- https://github.com/katef/libfsm/blob/main/src/libfsm/shortest.c#L113
- https://github.com/katef/libfsm/blob/main/src/libfsm/minimise.c#L27

sfstewman: Hmm... you're right.  Many of those are quite expensive.

`EXPENSIVE_ASSERT()` would be a very clear way to write them, and would also let us have builds that can choose between costly assertion behavior, still having some assertion guard rails, and turning off assertions.

silentbicycle:
Known worst cases for this include:

1. A very large ring of states where there is only one end state, and everything has the same label to the next state.
2. 0 -> 1a 'a'; 0 -> 1b 'b'; 1a -> 2a 'a'; 1b -> 2b 'b'; ... and both going a very long way before they reach an end state (a < shape instead of a ring, but otherwise the same issue).
3. A graph where there are a lot of states, but each only has a single edge, and they use as many different edge labels as possible.

The first two cases cause problems because they delay propagating info about different equivalence classes as much as possible -- the proof that states are observably different (because, for a given label, one leads to a different EC than the other) would only propagate one state at a time, so a ring with 1000 states would split to (999, 1), then (998, 1, 1), then (997, 1, 1, 1), ... and take quadratic work in the number of states. Both of these cases should be eliminated by a cleanup pre-pass that will appear on a subsequent PR.

The third case is actually a much bigger problem for Hopcroft, which explicitly tracks work remaining (a queue of splitters), and any time it makes progress, it needs to add a pass to the queue _for each label_. This maximizes the number of labels, but also minimizes the progress that can be made per pass. We avoid most of the overhead here because 1. we will partition on every label that makes progress, rather than doing a pass looking for cases where a particular splitter could (we are more opportunistic about propagating EC information), and 2. when the ECs are smaller than a configurable size threshold, instead of trying every label (when nearly all of them lead to the dead state), we check which edge label(s) are actually present in the EC and then check those. That gives us the option to do processing in terms of the number of edges in the EC's states, rather than the number of edge labels in the entire graph, and once the EC's state count is sufficiently small that should eliminate a lot of useless work.

src/libfsm/consolidate.c
```diff
@@ -0,0 +1,168 @@
...
+			if (!is_end) {
+				continue;
+			}
+			assert(fsm_isend(src, src_i));
+
+			/* Find the previous state that also maps to this
+			 * state, to carry the opaque from it. This could
+			 * become expensive when there are a very high number
+			 * of states, so it may be worth keeping a temporary
+			 * second mapping for just the end states. */
```
katef, line 132: I was wondering about doing this in `struct fsm` itself, too. As the primary representation for end states, I mean.

src/libfsm/minimise.c
```diff
@@ -7,49 +7,660 @@
...
+
 int
 fsm_minimise(struct fsm *fsm)
 {
-	int r;
+	int r = 0;
+	struct fsm *dst = NULL;
 
-	assert(fsm != NULL);
+	unsigned char labels[256];
```
katef, line 44: if this is for the number of symbols in the alphabet, there's a #define for this, `FSM_SIGMA_COUNT`
(and `FSM_SIGMA_MAX` for the maximum value of a symbol)

src/libfsm/minimise.c
```diff
@@ -7,49 +7,660 @@
...
+	for (id = 0; id < fsm->statecount; id++) {
+		struct fsm_edge e;
+		struct edge_iter ei;
+		unsigned char label;
+		for (edge_set_reset(fsm->states[id].edges, &ei);
+		     edge_set_next(&ei, &e); ) {
+			assert(e.state < fsm->statecount);
+			label = e.symbol;
+
+			if (label_set[label/64] & (1UL << (label & 63))) {
```
katef, line 149: Where does 64 come from? is this `sizeof (fsm_state_t) * CHAR_BIT`?

src/libfsm/minimise.c
```diff
@@ -7,49 +7,660 @@
...
+
+		li->has_labels_for_EC = 1;
+		li->i = 0;
+		li->limit = 0;
+		i = 0;
+		while (i < 256) {
+			if ((i & 31) == 0 && label_set[i/32] == 0) {
+				i += 32;
+			} else {
+				if (label_set[i/32] & ((uint32_t)1 << (i & 31))) {
```
katef, line 534: Same for the constants here.

src/libfsm/minimise_internal.h
```diff
@@ -0,0 +1,146 @@
...
+	 * Skipping the ECs after done_ec_offset improves performance
+	 * signifcantly, with far less complexity or overhead compared
+	 * to the splitter queue management in Hopcroft's algorithm
+	 * (which becomes more expensive as label_count increases). */
+	fsm_state_t done_ec_offset;
+
+	/* The set of labels that appear through the entire DFA. */
+	const unsigned char *dfa_labels;
+	size_t dfa_label_count;
+};
```
katef, line 66: The commenting all throughout this is excellent, thank you.

src/libfsm/minimise.c
silentbicycle: It's from the `uint64_t label_set[4]` defined above -- it's a bitset of FSM_SIGMA_COUNT entries. I'll change the type to `uint64_t label_set[FSM_SIGMA_COUNT/64]`, but I'm not sure what sort of note on the `/64` and `&63` part would make it clearer without visually obscuring the pattern. Any suggestions?

src/libfsm/minimise.c
silentbicycle: Commented in bf06b2ff.

src/libfsm/minimise.c
silentbicycle: Commented in bf06b2ff.

src/libfsm/minimise.c
silentbicycle: Changed in 45a445c3.

src/libfsm/minimise.c
katef: Ok, that works for me :)

PR #235, opened 2020-08-13
determinise: If there's no start state, just leave it empty.

silentbicycle:
Cherry-picked from #233, since it's tangential to that PR and easier to merge first.


katef:
Our idea here is that we're calling an empty FSM deterministic. With the definition that deterministic means there are no ambiguous edges. An empty FSM of course has no edges, so cannot be non-deterministic by that definition.

PR #236, opened 2020-08-18
parser.act: Use a hash table for token interning in the parser.

silentbicycle:
The parser's token interning dominates everything else when profiling
large test data sets, because previously it was walking an entire linked
list of tokens to find matches, which is quadratic on input size.

Switch to an add-only hash table with separate chaining for the tokens,
and grow whenever any of the chains are longer than MAX_CHAIN_LENGTH.

This uses FNV-1a for hashing, which is in the public domain.

I will update the generated parser code in a separate commit.

### DIFF
--- a/src/fsm/parser.act
+++ b/src/fsm/parser.act
@@ -44,21 +44,30 @@
 	typedef char *      string;
 	typedef fsm_state_t state;
 
+	struct act_statelist {
+		char *id;
+		fsm_state_t state;
+		struct act_statelist *next;
+	};
+
+	/* hash table with chaining */
+	#define DEF_BUCKETS 64	    /* must be a power of 2 */
+	#define MAX_CHAIN_LENGTH 16 /* grow hash table if longer than this */
+	struct act_stateset {
+		size_t longest_chain_length;
+		size_t bucket_count;
+		struct act_statelist **buckets;
+	};
+
 	struct act_state {
 		int lex_tok;
 		int lex_tok_save;
-		struct act_statelist *sl;
+		struct act_stateset states;
 	};
 
 	struct lex_state {
 		struct lx lx;
 		struct lx_dynbuf buf;
-	};
-
-	struct act_statelist {
-		char *id;
-		fsm_state_t state;
-		struct act_statelist *next;
 	};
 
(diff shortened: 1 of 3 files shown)


src/fsm/parser.act
```diff
@@ -44,23 +44,32 @@
 	typedef char *      string;
 	typedef fsm_state_t state;
 
+	struct act_statelist {
+		char *id;
+		fsm_state_t state;
+		struct act_statelist *next;
+	};
```
katef, line 51: I wonder if we shouldn't get rid of the idea of a string `id` for the fsm(1) syntax entirely. But maybe not, writing `1a`, `1b` etc is convenient for some tests, especially when the expected output is combined from two FSMs.

src/fsm/parser.act
```diff
@@ -90,6 +99,58 @@
...
+
+	static void
+	grow_state_hash_table(struct act_stateset *htab)
+	{
+		const size_t ncount = 2 * htab->bucket_count;
+		struct act_statelist **nbuckets = xcalloc(ncount, sizeof(*nbuckets));
+		size_t max_chain_length = 0;
+		size_t old_i;
+		unsigned mask = ncount - 1;
+		if (nbuckets == NULL) {
```
katef, line 127: This can never happen now (because xcalloc will exit)

src/fsm/parser.act
```diff
@@ -336,7 +413,10 @@
 		lx->clear      = lx_dynclear;
 		lx->free       = lx_dynfree;
 
-		act_state_s.sl = NULL;
+		act_state_s.states.buckets = calloc(DEF_BUCKETS, sizeof(act_state_s.states.buckets));
```
katef, line 416: `xcalloc` here too :)

src/fsm/parser.act
silentbicycle: I'd prefer to keep that flexible, it's been helpful when generating test data too.

src/fsm/parser.act
silentbicycle: Fixed, amended.

src/fsm/parser.act
silentbicycle: Fixed, amended.

PR #237, opened 2020-08-19
minimisation: Group initial EC partitions by shortest end distance

silentbicycle:
This adds a different version of `populate_initial_ecs` (enabled by `#define`ing `PARTITION_BY_END_STATE_DISTANCE` to `1`) which groups states with distinct end distances into their own ECs, rather than starting with two for final and non-final states. While this does not seem to have much impact on performance in typical cases (it may make `fsm_minimise` slightly slower), it completely eliminates the worst-case quadratic behavior in Moore's DFA minimization, which is when a large set of states with only a single deeply nested end state is partitioned, and info about the connection to the end state can only propagate one state further each pass. Now each state progressively further from the end state will be placed in a distinct EC to start.

The new implementation depends on an interface change in `fsm_trim` -- since it's already doing most of the necessary work while checking whether states have a path to any end states, add an argument to optionally track the shortest distance to an end state. Then, use this information in `populate_initial_ecs` to group states by shortest end distances, sorted by largest to smallest groups. Then, find the first EC with less than 2 states (if any); any ECs after that point are already done with processing. Hand the rest off to our variant of Moore's for the remaining processing.

Also, add an example of the previous worst case (`dfa_minimise_ring`) to `fuzz_minimise.c`'s regression tests, a ring of N states with one end state, with N set to 2000. This test still takes a while to run, because it's comparing the result against the Myhill-Nerode theorem-based minimisation test oracle (which is easy to verify but scales poorly), but the production implementation of minimisation should be a tiny portion of the runtime: on my laptop, `fsm_minimise` with `PARTITION_BY_END_STATE_DISTANCE` off takes about 34 msec, with it on the time drops to 0.15 msec.

### DIFF
--- a/include/fsm/fsm.h
+++ b/include/fsm/fsm.h
@@ -256,6 +256,15 @@ fsm_mergestates(struct fsm *fsm, fsm_state_t a, fsm_state_t b,
  * the state state's connected component), and optionally non-end
  * states that do not have a path to an end state.
  *
+ * If mode == FSM_TRIM_START_AND_END_REACHABLE and shortest_end_depth is
+ * non-NULL, then allocate and write an array with the length of the
+ * shortest distance to an end state for each state (post-trim) into
+ * *shortest_end_distance. Since checking for paths to an end state
+ * already does most of the work, this can be calculated at the same
+ * time with little overhead, and some operations (such as minimising)
+ * can make use of that information. On error, this array will
+ * automatically be freed.
+ *
  * Returns how many states were removed, or -1 on error.
  */
 enum fsm_trim_mode {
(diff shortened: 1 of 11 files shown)


include/fsm/fsm.h
```diff
@@ -256,6 +256,15 @@ fsm_mergestates(struct fsm *fsm, fsm_state_t a, fsm_state_t b,
...
  *
+ * If mode == FSM_TRIM_START_AND_END_REACHABLE and shortest_end_depth is
+ * non-NULL, then allocate and write an array with the length of the
+ * shortest distance to an end state for each state (post-trim) into
+ * *shortest_end_distance. Since checking for paths to an end state
+ * already does most of the work, this can be calculated at the same
+ * time with little overhead, and some operations (such as minimising)
+ * can make use of that information. On error, this array will
+ * automatically be freed.
+ *
```
katef, line 267: :+1:

src/libfsm/minimise.c
```diff
@@ -368,10 +382,255 @@ dump_ecs(FILE *f, const struct min_env *env)
 #endif
 }
 
-static void
-populate_initial_ecs(struct min_env *env, const struct fsm *fsm)
+#define PARTITION_BY_END_STATE_DISTANCE 1
```
katef, line 385: What's the plan to eventually resolve this?

silentbicycle: We could just remove it later. I wanted to leave it easy to enable/disable for benchmarking, particularly as we try to formalize the performance impact.

PR #242, opened 2020-09-06
Do not assume long has 64 bits.

hvdijk:
In hashset.c, siphash is called with a size of sizeof(unsigned long),
but siphash asserts that the size is either 8 or 16. Change h's type to
uint64_t instead. On platforms where long does not have 64 bits, this
will implicitly truncate the result in the return statement, which is
fine.

In minimise.c, use UINT64_C(1) instead of 1UL when a 64-bit constant is
needed.

This allows libfsm to work on 32 bit platforms.

### DIFF
--- a/src/adt/hashset.c
+++ b/src/adt/hashset.c
@@ -7,6 +7,7 @@
 #include <assert.h>
 #include <string.h>
 #include <stdio.h>
+#include <stdint.h>
 
 #include <adt/alloc.h>
 
@@ -24,7 +25,7 @@ static const unsigned char hashk[] = {
 unsigned long
 hashrec(const void *p, size_t n)
 {
-	unsigned long h = 0;
+	uint64_t h = 0;
 
 	siphash(p, n, hashk, (void *) &h, sizeof h);
 
--- a/src/libfsm/Makefile
+++ b/src/libfsm/Makefile
@@ -41,7 +41,7 @@ SRC += src/libfsm/intersect.c
 SRC += src/libfsm/subtract.c
 SRC += src/libfsm/walk2.c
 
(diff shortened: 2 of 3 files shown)


dhobsd:
Since uint64_t is optional, should we do something differently altogether
on such systems?

hvdijk:
> Since uint64_t is optional, should we do something differently altogether on such systems?

libfsm already uses `uint64_t` elsewhere, so this PR does not introduce any new portability issues.

libfsm also assumes C99, which requires integer types that are at least 64 bits wide. `uint64_t` may be optional, but `uint_least64_t` is not. It would be possible to use that instead, but using that consistently throughout the whole project would be a rather large undertaking.

silentbicycle:
This codebase is mostly C89, but some of the recent development uses C99. We (katef and I) have discussed switching over to C99 in general, but if so it's likely to happen after some other major changes (following replacing minimization) are complete.

There should be other places where I used an explicit cast to `(uint64_t)` rather than `1UL`, for that reason. Good catch there.

hvdijk:
@katef suggested to add `-std=c99` for these. There are a *lot* of files that already include `<stdint.h>` without specifying `-std=c99`: by forcing a warning to be generated for that and counting the warnings, I got 498. However, while using `<stdint.h>` in C90 mode does not normally result in any warnings, using `UINT64_C` does: it triggers "warning: use of C99 long long integer constant". I have updated the PR to add `-std=c99` to avoid that warning.

katef:
> Since uint64_t is optional, should we do something differently altogether on such systems?

About this part, I think dho means because these types are optional within C99. I'm fine with that; I've no intention of porting libfsm to an implementation of C99 that does not provide uint64_t :)

katef (approved):
Thank you!!

Issue #200, opened 2020-06-13
PCRE dialect doesn't handle option changes correctly


sfstewman:
Running `retest` to develop dfavm, I found a few inconsistencies between libfsm and the PCRE spec:

```[   1/  12] (pcre-tests.tst:594) regexp /(a(?i)bc|BB)x/ should match "bbx" but doesn't
[   2/  12] (pcre-tests.tst:606) regexp /^([ab](?i)[cd]|[ef])/ should match "Europe " but doesn't
[   3/  12] (pcre-tests.tst:608) regexp /^([ab](?i)[cd]|[ef])/ should match "France" but doesn't
[   6/  12] (pcre-tests.tst:1043) regexp /(?:(?i)a)b/ should not match "aB" but does
[   7/  12] (pcre-tests.tst:1139) regexp /((?i)AB(?-i)C|D)E/ should not match "abCe  " but does
[   8/  12] (pcre-tests.tst:1140) regexp /((?i)AB(?-i)C|D)E/ should not match "dE" but does
[   9/  12] (pcre-tests.tst:1141) regexp /((?i)AB(?-i)C|D)E/ should not match "De    " but does
```

(There are a few other unrelated errors.)

These errors fall into two groups:

1. Errors 1, 2, 3, 9: An option changed within a group applies to the remaining patterns and ALTs within that group
2. Errors 6, 7, 8,  An option changed within a group is still applied outside of that group

The relevant spec is here: http://www.pcre.org/current/doc/html/pcre2pattern.html#SEC13

> When one of these option changes occurs at top level (that is, not inside group parentheses), the change applies to the remainder of the pattern that follows. An option change within a group (see below for a description of groups) affects only that part of the group that follows it, so
>     (a(?i)b)c
> matches abc and aBc and no other strings (assuming PCRE2_CASELESS is not used). By this means, options can be made to have different settings in different parts of the pattern. Any changes made in one alternative do carry on into subsequent branches within the same group. 
> For example,
>       (a(?i)b|c)
> matches "ab", "aB", "c", and "C", even though when matching "C" the first branch is abandoned before the option setting.

Notes:
[...]

PR #244, opened 2020-09-07
Fix the handling of PCRE flags.

hvdijk:
This removes the `AST_EXPR_FLAGS` node type. This node type was used to indicate that the following nodes should be interpreted according to its flags, but "the following" was implemented as "the following children of its parent", rather than "the following elements in the original syntax until a closing parenthesis". "The following" was affected by internal re-ordering of nodes in alternatives, and "until a closing parenthesis" could not be determined anyway as non-capturing groups do not get their own nodes.

This commit removes the `AST_EXPR_FLAGS` node entirely and instead modifies the parser to pass the current `re_flags` to all AST construction functions and stores it in a new field in `struct ast_expr`. The compilation in `src/libre/ast_compile.c` is updated to look at this field to find whether `RE_ICASE` is enabled (the only flag currently implemented).

This commit updates the printing of ASTs as trees, dotfiles and PCREs to reconstruct the flags. As they are no longer represented as nodes, they do not get printed as nodes.

This fixes the option handling issues of #200.

Does this look like an okay approach, or would something completely different be better?


src/libre/parser.act
```diff
@@ -659,28 +654,28 @@
 	 */
 
 	<ast-make-empty>: () -> (node :ast_expr) = @{
-		@node = ast_make_expr_empty();
+		@node = ast_make_expr_empty(*flags);
```
katef, line 657: Noting to myself that `*flags` comes from a `%persistents%` section

src/libre/parser.act
```diff
@@ -90,11 +90,6 @@
 	#include "../comp.h"
 	#include "../../class.h"
 
-	struct flags {
-		enum re_flags flags;
-		struct flags *parent;
-	};
-
```
katef, line 97: So if we're setting these globally, and don't keep a stack or a tree, what's happening for cases like this?
```
; echo abcd | pcregrep '((?i:ABC))d'
abcd
; echo abcd | pcregrep '((?i:ABC))D'
;
```

I think we still need the idea of scope there because the flags do get restored outside of the capturing group.

Answer: I see you're saving and restoring the flags in pcre's parser.sid. That makes sense. So we're using the parser's call stack as our mechanism for scope.

This behaviour seems to apply to non-capturing groups, too. But we no longer need an AST node to hold that, because these are now set for every node. Okay! Great.

katef:
AST now:

image

katef:
The force-push above also removes `struct flags` entirely (with its `.pos` and `.neg` fields), leaving us now with just:
```diff
-	typedef struct flags *flags;
+	typedef enum re_flags * flags;
```
So now all flags are phrased in the positive in the AST, the idea of negative flags is now a detail of syntax only.

katef (approved):

hvdijk:
I found that with this PR, the ast printing of `\` and `|` was broken. These were not handled in `dot_escputc_html_record`, but it did not cause problems before as they only sometimes require escaping. I have updated the PR to always escape them.

silentbicycle (commented):
This looks good to me. I requested adding a test for the `?i-i` edge case, and there a few other potential test inputs on the PCRE pattern spec page (https://pcre.org/original/doc/html/pcrepattern.html#SEC13), but in the next couple weeks I also plan on doing comparative testing against pcregrep with generated regexes (using the subset of behavior we expect libfsm to handle), so that will shake out other unknown edge cases we need to test.

src/libre/parser.act
```diff
@@ -705,35 +700,45 @@
...
+		@flags = *flags;
+	@};
+
+	<ast-set-re-flags>: (flags :re_flags) -> () = @{
+		*flags = @flags;
+	@};
+
+	<ast-mask-re-flags>: (pos :re_flags, neg :re_flags) -> () = @{
+		/*
+		 * Note: in cases like `(?i-i)`, the negative is
```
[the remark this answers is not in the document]
hvdijk: Sure. I added `tests/pcre-flags/in5.re` containing `(?i-i)a`.

katef:
We'll need to handle the new flags field during AST rewriting, if we introduce any new nodes. Looking at ast_rewrite.c at the moment, I don't think we currently construct any new nodes, only removing existing ones.

katef:
Thank you again!

PR #248, opened 2020-09-09
Fix a memory leak when freeing AST_EXPR_ALT nodes

hvdijk:
Like AST_EXPR_CONCAT nodes, AST_EXPR_ALT nodes consist of two blocks of memory: the node itself, and the array of children. Free both of them.


katef (approved):
Ah!

PR #249, opened 2020-09-10
Fix printing of ASTs as PCRE.

hvdijk:
This fixes three aspects of PCRE printing:
- Not escaping two characters that do require it: { and }
- Not handling certain node types: (?:ab)* would crash
- Not taking precedence into account: a(?:b|c) would be printed as ab|c

The first of these is trivial, the second and third are related and a
bit more involved. It would have been easy to fix the crash by
supporting all node types in atomic(), and to handle precendence by
always emitting (?:...) where possibly required, but to avoid cluttering
the output this commit takes care to only use (?:...) grouping where
required, by ordering the node types according to PCRE precedence.

### DIFF
--- a/src/libre/print/pcre.c
+++ b/src/libre/print/pcre.c
@@ -21,19 +21,34 @@
 #include "../print.h"
 
 static int
-atomic(struct ast_expr *n)
+prec(struct ast_expr *n)
 {
 	switch (n->type) {
-	case AST_EXPR_EMPTY:
 	case AST_EXPR_LITERAL:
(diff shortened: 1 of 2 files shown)


src/libre/print/pcre.c
```diff
@@ -21,19 +21,34 @@
 #include "../print.h"
 
 static int
-atomic(struct ast_expr *n)
+prec(struct ast_expr *n)
```
katef, line 24: Good idea, this is a nice way to do it!

katef (approved):

Issue #251, opened 2020-09-11
libre leaks memory if certain errors are encountered during parsing


sfstewman:
When the `sid` parser encounters an error, it does not appear to free any partially-constructed pieces of the AST tree.

I found this working with `cvtpcre` with ASAN enabled, but it's easily triggered with `re` if ASAN is enabled:
```
% ./build/bin/re -r pcre '^\ca\cA\c[;\c:'
/^\ca\cA\c[;\c:/:2: Syntax error: unsupported operator

=================================================================
==21525==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0x7f44a8892dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x559d6fbd13b4 in ast_make_expr_alt src/libre/ast.c:419
    #2 0x559d6fbb475f in p_expr src/libre/parser.act:671
    #3 0x559d6fbb89e7 in p_re__pcre src/libre/dialect/pcre/parser.c:3326
    #4 0x559d6fbbb27f in parse_re_pcre src/libre/parser.act:909
    #5 0x559d6fbcdeba in re_parse src/libre/re.c:111
    #6 0x559d6fbce156 in re_comp src/libre/re.c:154
    #7 0x559d6fb20de5 in main src/re/main.c:688
    #8 0x7f44a85ba0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)

Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0x7f44a8892dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x559d6fbd0fd6 in ast_make_expr_concat src/libre/ast.c:371
    #2 0x559d6fbba396 in p_expr_C_Calt src/libre/parser.act:664
    #3 0x559d6fbb7716 in p_expr_C_Clist_Hof_Halts src/libre/dialect/pcre/parser.c:3116
    #4 0x559d6fbb478d in p_expr src/libre/dialect/pcre/parser.c:2558
    #5 0x559d6fbb89e7 in p_re__pcre src/libre/dialect/pcre/parser.c:3326
    #6 0x559d6fbbb27f in parse_re_pcre src/libre/parser.act:909
    #7 0x559d6fbcdeba in re_parse src/libre/re.c:111
[...]

PR #252, opened 2020-09-11
Changes AST parsing to use a "pooled" allocator

sfstewman:
AST nodes are allocated out of a "pool" which is basically a bump
allocator.  If an error is encountered in parsing, the entire pool is
released.

When the AST is freed, the entire pool is also released.

This version relies on a global pool variable that's later associated
with the `struct ast`, which makes parsing not thread-safe.

This is one potential solution to #251

### DIFF
--- a/src/libre/ast.c
+++ b/src/libre/ast.c
@@ -24,9 +24,6 @@
 static struct ast_expr the_tombstone;
 struct ast_expr *ast_expr_tombstone = &the_tombstone;
 
-/* FIXME: this isn't safe for multiple threads! */
-static struct ast_expr_pool *global_pool = NULL;
-
 struct ast_expr *
 ast_expr_pool_new(struct ast_expr_pool **poolp) 
 {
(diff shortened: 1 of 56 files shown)


katef:
#253 is based on this branch

src/libre/ast.c
```diff
@@ -24,11 +24,53 @@
 static struct ast_expr the_tombstone;
 struct ast_expr *ast_expr_tombstone = &the_tombstone;
 
+/* FIXME: this isn't safe for multiple threads! */
+static struct ast_expr_pool *global_pool = NULL;
```
katef, line 28: I think this is the only thing keeping me from merging this, at the moment.

src/libre/ast.c
```diff
@@ -24,26 +28,84 @@
...
+
+	p = *poolp;
+	if (p == NULL || p->count >= AST_EXPR_POOL_SIZE) {
+		p = malloc(sizeof *p);
+		if (p == NULL) {
+			return NULL;
+		}
+
+#if defined(ASAN)
+		ASAN_POISON_MEMORY_REGION(&p->pool, sizeof p->pool);
```
katef, line 48: this is very cool

katef (approved):
:+1: from me, thank you kindly

PR #253, opened 2020-09-13
Lots of fixes to various memory leaks

sfstewman:
So... I made the mistake of compiling with ASAN on Linux (running under WSLv2, but still Linux) and ... wow.  LSAN detects a ton of memory leaks in the standard test suite.

Here's an attempt to plug them, stacking on #252.

This isn't quite ready to merge yet.  There are some issues to resolve with #252 first.  But given the breadth of changes here, I wanted to raise them earlier rather than later.

### DIFF
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -349,6 +349,15 @@ gen_words(FILE *f, const struct fsm *fsm)
 }
 #endif /* 0 */
 
+static struct fsm *fsm_to_cleanup = NULL;
+
+static void
+do_fsm_cleanup(void)
+{
+	if (fsm_to_cleanup != NULL) {
+		fsm_free(fsm_to_cleanup);
+	}
+}
 
 int
 main(int argc, char *argv[])
@@ -364,6 +373,8 @@ main(int argc, char *argv[])
 	int (*query)(const struct fsm *, fsm_state_t);
 	int (*walk )(const struct fsm *,
 		 int (*)(const struct fsm *, fsm_state_t));
+
+	atexit(do_fsm_cleanup);
 
 	opt.comments = 1;
 	opt.io       = FSM_IO_GETC;
@@ -523,6 +534,8 @@ main(int argc, char *argv[])
 			q = NULL;
 		}
 
+		fsm_to_cleanup = q;
+
 		if (-1 == clock_gettime(CLOCK_MONOTONIC, &post)) {
 			perror("clock_gettime");
 			exit(EXIT_FAILURE);
@@ -645,6 +658,7 @@ main(int argc, char *argv[])
 	}
 
 	fsm_free(fsm);
+	fsm_to_cleanup = NULL;
 
 	return r;
 }
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -356,6 +356,7 @@ do_fsm_cleanup(void)
 {
 	if (fsm_to_cleanup != NULL) {
 		fsm_free(fsm_to_cleanup);
+		fsm_to_cleanup = NULL;
 	}
 }
 
--- a/src/fsm/parser.act
+++ b/src/fsm/parser.act
@@ -427,6 +427,7 @@
 		ADVANCE_LEXER; /* XXX: what if the first token is unrecognised? */
 		p_fsm(new, lex_state, act_state);
 
+		free(act_state_s.states.buckets);
 		lx->free(lx->buf_opaque);
 
 		return new;
--- a/src/fsm/parser.c
+++ b/src/fsm/parser.c
@@ -1256,11 +1256,12 @@ ZL1:;
 		ADVANCE_LEXER; /* XXX: what if the first token is unrecognised? */
 		p_fsm(new, lex_state, act_state);
 
+		free(act_state_s.states.buckets);
 		lx->free(lx->buf_opaque);
 
 		return new;
 	}
 
-#line 1265 "src/fsm/parser.c"
+#line 1266 "src/fsm/parser.c"
 
 /* END OF FILE */
--- a/src/libfsm/minimise.c
+++ b/src/libfsm/minimise.c
@@ -72,7 +72,8 @@ fsm_minimise(struct fsm *fsm)
 	}
 
 	if (fsm->statecount == 0) {
-		return 1;	/* empty -- no-op */
+		r = 1;
+		goto cleanup;
 	}
 
 	TIME(tv_pre);
@@ -81,7 +82,8 @@ fsm_minimise(struct fsm *fsm)
 	LOG_TIME_DELTA("collect_labels");
 
 	if (label_count == 0) {
-		return 1;	/* no edges -- no-op */
+		r = 1;
+		goto cleanup;
 	}
 
 	mapping = f_malloc(fsm->opt->alloc,
--- a/src/libfsm/state.c
+++ b/src/libfsm/state.c
@@ -211,6 +211,9 @@ fsm_compact_states(struct fsm *fsm,
 	for (i = 0, dst = 0, removed_count = 0; i < orig_statecount; i++) {
 		assert(dst <= i);
 		if (mapping[i] == FSM_STATE_REMAP_NO_STATE) { /* dead */
+			state_set_free(fsm->states[i].epsilons);
+			edge_set_free(fsm->opt->alloc, fsm->states[i].edges);
+
 			fsm->statecount--;
 			removed_count++;
 		} else {				      /* keep */
--- a/src/libfsm/walk/reachable.c
+++ b/src/libfsm/walk/reachable.c
@@ -21,60 +21,71 @@ fsm_reachable(const struct fsm *fsm, fsm_state_t state,
 	int any,
 	int (*predicate)(const struct fsm *, fsm_state_t))
 {
-	struct dlist *list;
-	struct dlist *p;
+	int r, stopcond;
+
+	fsm_state_t *q;
+	size_t i, qtop;
 
 	assert(state < fsm->statecount);
 	assert(predicate != NULL);
+
+	for (i=0; i < fsm->statecount; i++) {
+		fsm->states[i].visited = 0;
+	}
+
+	q = f_malloc(fsm->opt->alloc, fsm->statecount * sizeof *q);
+	if (q == NULL) {
+		return -1;
+	}
 
 	/*
 	 * Iterative depth-first search.
 	 */
 	/* TODO: write in terms of fsm_walk or some common iteration callback */
 
-	list = NULL;
+	q[0] = state;
+	qtop = 1;
+	fsm->states[state].visited = 1;
 
-	if (!dlist_push(fsm->opt->alloc, &list, state)) {
-		return -1;
-	}
+	stopcond = (any) ? 1 : 0;
 
-	while (p = dlist_nextnotdone(list), p != NULL) {
+	while (qtop > 0) {
 		struct edge_iter it;
 		struct fsm_edge e;
+		fsm_state_t state;
+		int pred_result;
 
-		if (any) {
-			if (predicate(fsm, p->state)) {
-				dlist_free(fsm->opt->alloc, p);
-				return 1;
-			}
-		} else {
-			if (!predicate(fsm, p->state)) {
-				dlist_free(fsm->opt->alloc, p);
-				return 0;
-			}
+		assert(qtop > 0);
+
+		state = q[--qtop];
+
+		pred_result = !!predicate(fsm, state);
+		if (pred_result == stopcond) {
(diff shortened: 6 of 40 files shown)


src/fsm/main.c
```diff
@@ -645,6 +658,7 @@ main(int argc, char *argv[])
 	}
 
 	fsm_free(fsm);
+	fsm_to_cleanup = NULL;
```
BenBE, line 661: Why not use the atexit proc instead?

src/fsm/parser.act
```diff
@@ -427,6 +427,7 @@
 		ADVANCE_LEXER; /* XXX: what if the first token is unrecognised? */
 		p_fsm(new, lex_state, act_state);
 
+		free(act_state_s.states.buckets);
```
BenBE, line 430: It's preferable to NULL a pointer after freeing it to avoid use-after-free issues. The compiler will usually DSE those writes if it can prove they are unnecessary.

src/fsm/parser.c
```diff
@@ -1256,11 +1256,12 @@ ZL1:;
 		ADVANCE_LEXER; /* XXX: what if the first token is unrecognised? */
 		p_fsm(new, lex_state, act_state);
 
+		free(act_state_s.states.buckets);
```
BenBE, line 1259: Explicit NULL; avoid use-after-free potential.

src/libfsm/walk/reachable.c
```diff
@@ -21,60 +21,71 @@ fsm_reachable(const struct fsm *fsm, fsm_state_t state,
...
-	if (any) {
-		return 0;
-	} else {
-		return 1;
+cleanup:
+	for (i=0; i < fsm->statecount; i++) {
+		fsm->states[i].visited = 0;
 	}
+
+	f_free(fsm->opt->alloc, q);
```
BenBE, line 86: Explicit NULL; avoid use-after-free potential.

src/libre/ast.c
```diff
@@ -428,7 +492,6 @@ ast_make_expr_alt(enum re_flags re_flags)
 
 	res->u.alt.n = malloc(res->u.alt.alloc * sizeof *res->u.alt.n);
```
BenBE, line 493: Prefer `calloc` or check for integer overflow in argument expression.

src/libre/ast.c
```diff
@@ -617,7 +680,6 @@ ast_make_expr_named(enum re_flags re_flags, const struct class *class)
...
```
BenBE, line 681: Use `calloc` of check for integer overflow in argument expression.

src/libre/ast_rewrite.c
```diff
@@ -226,6 +228,12 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
 
 tombstone:
 
+	for (i = 0; i < n->u.concat.count; i++) {
+		ast_expr_free(n->u.concat.n[i]);
+	}
+
+	free(n->u.concat.n);
```
BenBE, line 235: Explicit NULL; avoid use-after-free potential.

src/re/main.c
```diff
@@ -476,6 +528,18 @@ endleaf_json(FILE *f, const void *state_opaque, const void *endleaf_opaque)
 	return 0;
 }
 
+static struct fsm *fsm_to_cleanup = NULL;
+
+static void
+do_fsm_cleanup(void)
+{
+	if (fsm_to_cleanup != NULL) {
+		fsm_free(fsm_to_cleanup);
```
BenBE, line 537: Explicit NULL; avoid use-after-free potential; even as this is the last code to run ;-) It's just cleaner to have it consistent as this symbol is globally visible.

tests/hashset/hashset4.c
```diff
@@ -59,16 +59,31 @@ int *next_int(int reset) {
 int main(void) {
 	struct hashset *s = hashset_create(NULL, hash_int, cmp_int);
 	size_t i;
+	int **plist;
+
+	plist = malloc(10000 * sizeof *plist);
```
BenBE, line 64: Perfer `calloc` to make array character of this block of memory more visible.

tests/hashset/hashset4.c
```diff
@@ -59,16 +59,31 @@ int *next_int(int reset) {
 int main(void) {
 	struct hashset *s = hashset_create(NULL, hash_int, cmp_int);
```
BenBE, line 60: Unchecked allocation?

tests/hashset/hashset5.c
```diff
@@ -53,13 +53,29 @@ int main(void) {
 	struct hashset *s = hashset_create(NULL, hash_int, cmp_int);
```
BenBE, line 53: Unchecked allocation?

tests/hashset/hashset5.c
```diff
@@ -53,13 +53,29 @@ int main(void) {
...
 	int a[3] = {1200,2400,3600};
 	size_t i;
+	int **plist;
+
+	plist = malloc(5000 * sizeof *plist);
```
BenBE, line 58: Prefer `calloc` to highlight array character of this allocation.

tests/set/set4.c
```diff
@@ -35,9 +35,23 @@ int *next_int(void) {
 int main(void) {
 	struct set *s = set_create(NULL, cmp_int);
 	size_t i;
+	int **plist;
+
+	plist = malloc(5000 * sizeof *plist);
```
BenBE, line 40: Prefer `calloc` for array allocations.

tests/set/set4.c
```diff
@@ -35,9 +35,23 @@ int *next_int(void) {
 int main(void) {
 	struct set *s = set_create(NULL, cmp_int);
```
BenBE, line 36: Unchecked allocation.

tests/set/set5.c
```diff
@@ -36,13 +36,27 @@ int main(void) {
 	struct set *s = set_create(NULL, cmp_int);
```
BenBE, line 36: Unchecked allocation.

tests/set/set5.c
```diff
@@ -36,13 +36,27 @@ int main(void) {
...
 	int a[3] = {1200,2400,3600};
 	size_t i;
+	int **plist;
+
+	plist = malloc(5000 * sizeof *plist);
```
BenBE, line 41: Prefer `calloc` for array allocations.

tests/set/set6.c
```diff
@@ -36,9 +36,16 @@ int main(void) {
 	struct set *s = set_create(NULL, cmp_int);
```
BenBE, line 36: Unchecked allocation.

tests/set/set6.c
```diff
@@ -36,9 +36,16 @@ int main(void) {
...
 	struct set_iter iter;
 	size_t i;
+	int **plist;
+
+	plist = malloc(5000 * sizeof *plist);
```
BenBE, line 41: Prefer `calloc` for array allocations.

BenBE (changes requested):
Annotated some missing allocation checks. Also several places where `calloc` is probably the more canonical choice to use. Lastly when a pointer inside a struct is `free`d while the struct is reused, the pointer should be clear'd immediately after (either explicit or implicit by overwriting the whole structure).

src/fsm/main.c
sfstewman: It's a good question.  I don't particularly like the `atexit()` code.  I added it to silence LSAN false positives and avoid adding a lot of cleanup code for all exit paths.  I'm not 100% sure that this should go into the actual code.

So... this is a bit of a style question for @katef and others (eg: @silentbicycle).

src/fsm/parser.act
sfstewman: That seems unnecessary here.  The function returns immediately after freeing this and another buffer.  `act_state_s` is a variable with automatic storage duration; if this pointer leaks outside, we have other problems.

src/fsm/parser.c
sfstewman: `parser.c` is generated from `parser.act`.

src/libfsm/walk/reachable.c
sfstewman: This is in a short section with the label `cleanup`.  Again, the next line is a return, and `q` is a variable with automatic storage duration.  There's no use-after-free issue.

src/libre/ast.c
sfstewman: This is probably a situation for `calloc`.

tests/set/set4.c
sfstewman: These are short test cases.  I can add an assert(), but ... if `set_create()` returns `NULL`, the test will fail with a segfault.

tests/set/set5.c
sfstewman: Same as `set4.c`

tests/set/set5.c
sfstewman: I disagree quite strongly with this preference.  Sometimes you want `calloc` and sometimes you don't want to initialize all bytes to zero.

In any case, the argument is a compile-time constant and the compiler should warn about overflow.

tests/set/set6.c
sfstewman: Same as `set4.c`

tests/set/set6.c
sfstewman: Same as `set5.c`

sfstewman:
Addressed the calloc() concern that @BenBE had in AST allocation.

Also improved some of the test code (`tests/hashset/*.c` and `tests/set/*.c`) in a few ways:

1. Always assert that the set is created
2. `plist` allocated with `calloc`.  In this case, it's almost certainly doesn't matter either way, but calloc is convenient.
3. Replaces hard-coded constants with enum values or sizeof-derived expressions.  This has been bugging me for a while.

tests/set/set6.c
```diff
@@ -32,31 +32,36 @@ int *next_int(void) {
...
 	struct set *s = set_create(NULL, cmp_int);
 	struct set_iter iter;
 	size_t i;
 	int **plist;
+	int *p;
+
+	assert(s != NULL);
 
-	plist = malloc(5000 * sizeof *plist);
+	plist = calloc(5000, sizeof *plist);
```
BenBE, line 46: Shouldn't this also become `COUNT`?

BenBE (approved):
LGTM.

src/libre/ast_rewrite.c
katef: Hi Ben! We don't usually set things to NULL after free in this codebase. Mostly because that hints that they'll be used again, and if they're being freed, they shouldn't be. Instead they go out of scope, or get assigned some other non-NULL value or something like that, depending on each situation.

tests/set/set4.c
katef: Hiya, we don't tend to calloc unless there's an explicit need to set items to all-bytes-zero. Pointer and size arithmetic is absolutely everywhere in C, I don't think it's fair to try to convey all array-like uses with calloc.

src/fsm/main.c
katef: Hiya, I'm fine with using atexit where we need to. In most cases I don't think we need to.

src/libre/ast_rewrite.c
BenBE: I'm totally fine with this too, if the codebase ensures that if it reuses such structures there is no risk of such invalid pointers being accessed (other than by pointing them to newly allocated memory). Assigning NULL usually is a defense-in-depth strategy to ensure exactly this and helps avoid security issues if such an invalid pointer is accidentally accessed.

src/libfsm/minimise.c
```diff
@@ -72,7 +72,8 @@ fsm_minimise(struct fsm *fsm)
 	}
 
 	if (fsm->statecount == 0) {
-		return 1;	/* empty -- no-op */
```
katef, line 75: Can we keep these comments please? same below

src/libfsm/state.c
```diff
@@ -211,6 +211,9 @@ fsm_compact_states(struct fsm *fsm,
 	for (i = 0, dst = 0, removed_count = 0; i < orig_statecount; i++) {
 		assert(dst <= i);
 		if (mapping[i] == FSM_STATE_REMAP_NO_STATE) { /* dead */
+			state_set_free(fsm->states[i].epsilons);
+			edge_set_free(fsm->opt->alloc, fsm->states[i].edges);
+
```
katef, line 216: Aha!!

katef:
I think this will need rebasing now #253 has gone in, because the common ancestor there changed.

sfstewman:
This has been rebased onto `main` in #265.  Closing.

PR #254, opened 2020-09-18
Move fsm parser into libfsm

hvdijk:
The fsm file format is already part of libfsm proper: the print functions are part of libfsm. This moves the parser for the fsm file format into libfsm as well, so that it can be reused by other utilities.

This is a prerequisite for the PR that will come later to build ASTs that correspond to a given FSM. It can be be merged on its own if so desiref, but does not need to be, I have created a separate branch and PR in advance for it for easier review.


katef (approved):
This makes sense. I had originally hoped to keep the .fsm syntax outside of the library proper (I don't particularly like the syntax),

I don't think that we generate it as output is good reason to include it - for example, we generate .dot and various other formats. But we need it from libre, so that'll do for me.

PR #256, opened 2020-09-18
Add fsm file support to re

hvdijk:
This commit provides an option to construct an AST from a FSM, and adds an option to re that uses this to load existing .fsm files. It can be used to construct ASTs and thus regular expressions for arbitrary FSMs.

### DIFF
--- a/src/libre/ast_new_from_fsm.c
+++ b/src/libre/ast_new_from_fsm.c
@@ -1,0 +1,498 @@
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <assert.h>
+
+#include <fsm/fsm.h>
+#include <fsm/options.h>
+#include <fsm/pred.h>
+#include <fsm/walk.h>
+
+#include <re/re.h>
+
+#include "ast.h"
+#include "ast_analysis.h"
+
+static int
+change_to_alt(struct ast_expr **n)
+{
+	struct ast_expr *alt;
+
+	assert(n);
+	assert(*n);
+
+	if ((*n)->type == AST_EXPR_ALT)
+		return 1;
+
+	alt = ast_make_expr_alt(0);
+	if (!alt)
+		return 0;
+
+	if (!ast_add_expr_alt(alt, *n)) {
+		ast_expr_free(alt);
+		return 0;
+	}
+
+	*n = alt;
+	return 1;
+}
+
+/*
+ * RE state/edge
+ *
+ * We clone the FSM into a custom structure that allows edges to be
+ * annotated by arbitrary AST expressions. The states double as the
+ * self-edges. Outgoing edges to another state are stored in an
+ * intrusive circular doubly linked list, ordered by state, using the
+ * succprev and succnext pointers. Incoming edges from another state
+ * are likewise accessible using the predprev and prednext pointers.
+ *
+ * expr is the expression associated with the transition. It can be null
+ * only for self-edges.
+ */
+
+struct rese
+{
+	struct rese *from, *to;
+	struct rese *predprev, *prednext;
+	struct rese *succprev, *succnext;
+	struct ast_expr *expr;
+};
+
+static struct rese *
+build_restates(unsigned int numstates)
+{
+	size_t numbytes;
+	struct rese *result, *cur, *end;
+
+	if (!numstates)
+		return NULL;
+
+	numbytes = numstates * sizeof *result;
+	if (numbytes / sizeof *result != numstates)
+		return NULL;
+
+	result = malloc(numbytes);
+	if (!result)
+		return NULL;
+
+	for (cur = result, end = &result[numstates]; cur != end; cur++) {
+		cur->from = cur;
+		cur->to = cur;
+		cur->succprev = cur;
+		cur->succnext = cur;
+		cur->predprev = cur;
+		cur->prednext = cur;
+		cur->expr = NULL;
+	}
+
+	return result;
+}
+
+static struct rese *
+build_reedge(struct rese *from, struct rese *to)
+{
+	struct rese *result, *succ, *pred;
+
+	assert(from);
+	assert(to);
+
+	if (from == to)
+		return from;
+
+	pred = to->prednext;
+	for (;;) {
+		assert(pred->to == to);
+		if (pred->from == from)
+			return pred;
+		if (pred->from == to || pred->from > from)
+			break;
+		pred = pred->prednext;
+	}
+
+	succ = from->succnext;
+	for (;;) {
+		assert(succ->from == from);
+		if (succ->to == to) {
+			assert(!"should have caught edge in pred loop");
+			return succ;
+		}
+		if (succ->to == from || succ->to > to)
+			break;
+		succ = succ->succnext;
+	}
+
+	result = malloc(sizeof *result);
+	if (!result)
+		return NULL;
+
+	result->from = from;
+	result->to = to;
+	result->predprev = pred->predprev;
+	result->prednext = pred;
+	result->succprev = succ->succprev;
+	result->succnext = succ;
+	result->expr = NULL;
+	result->predprev->prednext = result;
+	result->prednext->predprev = result;
+	result->succprev->succnext = result;
+	result->succnext->succprev = result;
+	return result;
+}
+
+static int
+build_edge_expr(const struct fsm *fsm, fsm_state_t from, fsm_state_t to, struct ast_expr *expr, void *opaque)
+{
+	struct rese *restates, *reedge;
+
+	assert(fsm);
+	assert(expr);
+	assert(opaque);
+
+	restates = opaque;
+
+	reedge = build_reedge(&restates[from], &restates[to]);
+	if (!reedge)
+		return 0;
+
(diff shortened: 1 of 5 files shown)


src/re/main.c
```diff
@@ -672,12 +696,20 @@ main(int argc, char *argv[])
 				break;
 			}
 
-			if (yfiles) {
+			if (fsmfiles || yfiles) {
```
katef, line 699: I'm amazed you managed to navigate this mess!

src/libre/ast_new_from_fsm.c
```diff
@@ -0,0 +1,474 @@
...
+	struct fsm_options opt;
+
+	unsigned int numstates;
+	fsm_state_t start;
+	struct rese *restates;
+	struct rese *restart, *reend;
+
+	unsigned int i;
+	struct rese *state;
+
```
katef, line 374: Could you assert() on the contractual constraints here please? and comment them in the header.
Or maybe there are none, I'm not sure!

src/libre/ast_new_from_fsm.c
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	ast->expr = expr;
+
+	if (!ast_rewrite(ast, 0)) {
+		ast_free(ast);
+		return NULL;
+	}
+
+	return ast;
```
katef, line 473: Testing this function with some random FSMs (from `fuzz/gengraph $RANDOM 10 5 5 abcd 0`) I found it's very easy to get `ast_new_from_fsm` to fail. That's fine, if it's for a reason that makes sense, but I can't work out what it is! Am I missing something? I tried with DFA, too.

Can you set `errno` for the cases where we return failure here please? `EINVAL` especially. Currently we indicate failure as "ast_new_from_fsm: Success", which should never be visible.

katef (commented):
This is really good work, thank you.

I commented a bit about some smaller things. I have nothing useful to say about the algorithm itself!

src/libre/ast_new_from_fsm.c
```diff
@@ -0,0 +1,474 @@
...
+	struct rese *restates;
+	struct rese *restart, *reend;
+
+	unsigned int i;
+	struct rese *state;
+
+	numstates = fsm_countstates(fsm);
+
+	if (!fsm_getstart(fsm, &start))
+		return 0;
```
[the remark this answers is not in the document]
hvdijk: Done!

src/libre/ast_new_from_fsm.c
hvdijk: It should work for arbitrary FSMs, even though apparently it does not yet :) I added an assert that `fsm` must not be a null pointer, I think you typically leave that one out of any header comments, but I can add it if you like.

src/libre/ast_new_from_fsm.c
hvdijk: We should only return failure if another function failed, in which case that function should have set `errno`. I see now that I'm returning `NULL` for FSMs without a start state. I shouldn't be doing that, I should be returning `AST_EXPR_TOMBSTONE`: it's a valid FSM but can never match anything. There may be more cases like that, I'll check!

hvdijk: Fixed the ones I saw, and then the fuzzer found something that fails in `ast_rewrite` but is also reproducible from a straight PCRE regex, will look into that and put that in a separate PR if I find anything!

src/libre/ast_new_from_fsm.c
katef: Yes, no need to comment that :)

I think we must assume that there's a reachable end state perhaps?

katef: Oh no, we'd accept that and produce a tombstone. nice

src/libre/ast_new_from_fsm.c
katef: Brilliant!

katef: I can still produce `ast_new_from_fsm: Success` for this entertaining-looking FSM:
image

I suspected because no end state is reachable. But this works:
```
; echo 'start: 0;' | ./build/bin/re -pl pcre -f -
```

I trimmed down a test case, and can reproduce it with this:
image

```
0 -> 1 'a';

start: 2; # same error if this is 1 instead
end: 0;
```

hvdijk: Huh, that is surprising. That exact test case does work here:
```
$ cat test.fsm
0 -> 1 'a';

start: 2; # same error if this is 1 instead
end: 0;
$ build/bin/re -f -pl pcre test.fsm
[^\x00-\xff]
$
```

hvdijk: After a rebuild with optimizations enabled, I do get the result you are seeing. Will try to figure out what is going on.

hvdijk: Ah, the risks of compiling without warnings! There was a missing `return` statement in `remove_state`. In non-optimized builds, I happened to get a non-zero return value that indicates success, in optimized builds I (and you) happened to get a zero return value that indicates an error. That is fixed now.

katef: Also visible for this case:
image

hvdijk: Yeah, it would have happened for many cases. To clarify: are you still seeing this now? I pushed a fix for this, and on the earlier commit, I was able to get that same (non-)error for `a?b` (only with optimizations enabled), on the latest it works, with or without optimizations.

Looking more closely at the output of the compiler, there was also one place where I was using a C99 feature, a declaration after a statement. I will push a fix for that too.

katef: You know you can compile with warnings too! `bmake -r CC=gcc` should do it.

katef: Confirmed that all these case work for me now, thank you!

katef:
This is very cool
```
; ./build/bin/re -bp '[a-z][a-z0-9_]+' > /tmp/ident.fsm                                     
; ./build/bin/re -bp 'cheese' > /tmp/cheese.fsm                                    
; ./build/bin/fsm -pt subtract /tmp/ident.fsm /tmp/cheese.fsm | ./build/bin/re -bpl pcre -f -
(?-)(?:a|b|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*|c(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*|ch(?:|(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*)|che(?:|(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*)|chee(?:|(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*)|chees(?:|(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*)|cheese(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)(?:0|1|2|3|4|5|6|7|8|9|_|a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w|x|y|z)*
```

katef (approved):

PR #257, opened 2020-09-19
Fix memory error in rewrite_alt.

hvdijk:
n->u.alt.alloc is the number of nodes that n->u.alt.n can hold, but was being assigned the number of bytes. This leads later code to conclude enough storage is already available and to write beyond the end of n->u.alt.n.


katef (approved):
Good find. Thank you!

PR #258, opened 2020-09-20
Fix two issues in PCRE printing.

hvdijk:
Fix unused local variable `localflags`: the idea of PR #249 was to use this variable in the call to `pp_iter`, but this was accidentally still using `re_flags`. This had no effect because we never need to group an operand of `|`, as `|` has the lowest precedence of all operators, but rather than simply remove the code, it is preserved to allow grouping operands of `|` in the future for simpler flag printing, as in `a|(?i:b)|c*`, which currently gets printed as `a|(?i)b|(?-i)c*`.

Fix printing of changed flags to ignore a change in global flags: `RE_ANCHORED` applies to whole regexps only, not individual
subexpressions, and cannot be turned on or off. This cannot occur when constructing ASTs from regexps, as the `RE_ANCHORED` flag will be the same for all AST nodes, but can happen when constructing ASTs from FSMs when additionally using the `-b` option, which should have no effect as FSMs are implicitly anchored.

### DIFF
--- a/src/libre/print/pcre.c
+++ b/src/libre/print/pcre.c
@@ -67,6 +67,10 @@ re_flags_print(FILE *f, enum re_flags fl)
 	if (fl & RE_ZONE   ) { fprintf(f, "z"); }
 }
 
+enum {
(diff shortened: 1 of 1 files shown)


katef (approved):
Thanks!

PR #259, opened 2020-09-21
Rewrite `(?:|xyz)` and `(?:xyz|)` to `(?:xyz)?`.

katef:
This helps a little for rendering out to more natural regexp syntax.


src/libre/ast_rewrite.c
```diff
@@ -317,6 +317,30 @@ rewrite_alt(struct ast_expr *n, enum re_flags flags)
 		return 1;
 	}
 
+	if (n->u.alt.count == 2) {
```
katef, line 320: @hvdijk points out that this would also apply for n > 2, which I hadn't thought of.

PR #260, opened 2020-09-21
More PCRE parser fixes: (?flags:expr) applying to end-of-pattern

hvdijk:
This extends the lx syntax to allow multiple zone entry and exit mappings, uses it to simplify the PCRE bracket handling by combining everything into a single zone, uses it to extend the PCRE bracket handling by adding -] as an exit mapping taking - as a literal, and uses it to implement support for (?flags:expr) to indicate that the flags are only in effect until the end of expr, equivalent to (?:(?flags)expr).


katef (approved):
The lx feature here is great work, I really appreciate it, and you've implemented it well.

I think the diff to the pcre lexer.lx speaks for itself :)

katef:
```
'[' -> $opengroup, '[^' -> $opengroupinv, '[]' -> $opengroupcb, '[^]' -> $opengroupinvcb .. ']' -> $closegroup, '-]' -> $closegrouprange { }
```
One day, I hope we can have lx syntax using anchors to split these into separate tokens. Something like:
```
'[' -> $opengroup .. ']' -> $closegroup {
    /\^/z -> $invert; # /z meaning a flag to make ^$ apply to the start and end of a zone
    ...
}

PR #262, opened 2020-09-30
Fix builds for glob, utf8dfa examples

nilium:
Convert a few uses of `struct fsm_state *` to `fsm_state_t` in the glob
and utf8dfa examples, particularly around `fsm_addstate`.

Adds an `fsm_determinise` call to the glob example. This is because
it'll still have epsilon edges without it after compiling, causing the
later `fsm_all(fsm, fsm_isdfa)` assertion to fail. This follows the
lead from commit 631c56f3's message. (This doesn't seem to be necessary
for utf8dfa since it only creates literal edges between states.)

Spotted the glob case while reimplementing it using Crystal bindings,
so updated that and utf8dfa (which needed similar changes). The iprange
example also looks like it needs an update but I haven't quite got it
working yet, so it's not part of this change.

### DIFF
--- a/examples/glob/main.c
+++ b/examples/glob/main.c
@@ -49,7 +49,7 @@ usage(void)
 static int
 match(const struct fsm *fsm, const char *s)
 {
-	const struct fsm_state *state;
+	fsm_state_t state;
 
 	assert(fsm != NULL);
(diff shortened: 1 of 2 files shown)


katef (approved):
Oh! Thank you!

PR #265, opened 2020-10-07
Lots of fixes to various memory leaks 

sfstewman:
This is a rebased version of #253, but pushed to the `katef` repo so github checks will actually run (*grumble*).

This is now ready to merge and plugs all of the leaks that LSAN finds in the unit tests.


katef (approved):
Awesome. Thank you so much for this, I appreciate it.

PR #266, opened 2020-10-08
Various PCRE syntax improvements

sfstewman:
Most of this was driven by a desire to make `cvtpcre` correctly handle parsing the PCRE test suite, and for it to correctly identify unsupported PCRE features so the output wouldn't have to be heavily hand-edited to eliminate things like backreferences.

### PCRE syntax:
- Handles `\x` codes without trailing hex digits
- Handles `\cX` control characters
- Handles `\b` and `\e` in character classes
- Handles PCRE comments `(?# ... )`
- Handles `\A` and `\z`
- Handles `\E` without a preceding `\Q`
- Handles partial support for named capture groups, although does not capture the name
- Repeats now follow PCRE rules, which are somewhat more strict

### Marks a lot of PCRE syntax as "unsupported" to give better error messages:
- Backreferences denote through `\1` .. `\7`, `\g`, and `\k`
- Lookahead/lookbehind assertions
- Non-greedy repeats (eg: `*?`) and atomic repeats (eg: `*+`) are marked as unsupported
- `\b`, `\B`, and `\G` are unsupported
- Various `(*VERB)`, `(*VERB:NAME)`, and `(*:NAME)` backtracking controls are unsupported
- `\K` "capture group reset" is (currently) unsupported
- `\Z` is (currently) unsupported

### Other related improvements:
- `fsm_example` outputs `\xHH` for unprintable characters
- Adds `M` records to `retest` to turn on/off regexp modifiers
- Fixes some bugs in `cvtpcre`
- `cvtpcre` optionally takes input file and output file arguments
- Improves how `cvtpcre` parses the PCRE test suites


sfstewman:
I'm not 100% sure that this is ready to merge.  I need to write some test cases for a bunch of the grammar that's now supported or is now marked as unsupported.

katef:
Am happy to merge this ahead of time, if the tests are coming later. Because this way around eases the work for merging, and then we don't end up blocked on this (excellent!) work, and also don't end up having to deal with merges on this side. I prefer merges to be handled wherever the conflicts are simpler.

katef:
I also think we have more theft tests coming, and we'll pick up more of the pcre equivalence there too, eventually.

include/re/re.h
```diff
@@ -85,7 +85,8 @@ enum re_errno {
 	RE_EFLAG        = 11 | RE_MARK,
 	RE_EXCLOSEFLAGS = 12 | RE_MARK,
 	RE_EXUNSUPPORTD = 13 | RE_MARK,
-	RE_EBADCP       = 14 | RE_MARK
+	RE_EBADCP       = 14 | RE_MARK,
+	RE_BADCOMMENT   = 15 | RE_MARK
```
katef, line 89: amazing that a regexp dialect has a notion of comments, let alone ill-formed ones

src/libre/dialect/pcre/lexer.lx
```diff
@@ -8,10 +8,47 @@
...
+# > at the start point of the matching process, as specified by the
+# > startoffset argument of pcre2_match(). It differs from \A when the
+# > value of startoffset is non-zero. By calling pcre2_match() multiple
+# > times with appropriate arguments, you can mimic Perl's /g option, and
+# > it is in this kind of implementation where \G can be useful. 
+'\G' -> $unsupported;
+
+# A lone '\E' without a preceding \Q is ignored
+'\E';
+
```
katef, line 51: Excellent, thank you

src/libre/dialect/pcre/lexer.lx
```diff
@@ -8,10 +8,47 @@
...
+# > and dollar (described in the next section) in that they only ever
+# > match at the very start and end of the subject string, whatever
+# > options are set. Thus, they are independent of multiline mode. 
+'\A' -> $start;
+'\z' -> $end;
+
+# not yet implemented:
+# > matches at the end of the subject
+# >    also matches before a newline at the end of the subject
+'\Z' -> $unsupported;
```
katef, line 36: Ah the unsupported token makes sense. I hadn't imagined that!

src/libre/dialect/pcre/lexer.lx
```diff
@@ -59,20 +97,97 @@
...
+'\0'                   -> $oct;
+
+# PCRE treats \1 ... \7 as back references outside of a character class
+#
+# It's actually more complicated than that.  \<digits> is treated as a
+# backref as long as there are _already_ that many capture groups in the
+# expression.  Given that back references are unsupported, this seems
+# like a overkill for a seldom-used feature.  However, it's nice to
+# indicate that the usual backrefs aren't supported.
+'\'   /[1-7]/          -> $unsupported;
```
katef, line 110: This works for me.

src/libre/dialect/pcre/lexer.lx
```diff
@@ -59,20 +97,97 @@
...
+	'+'      -> $unsupported;
+	/[0-9]+/ -> $unsupported;
+	'R'      -> $unsupported;
 }
 
+
+# recursion with named parentheses
+'(?&' -> $unsupported;
+'(P>' -> $unsupported;
+
```
katef, line 190: This dialect needs a prop 65 warning

src/libre/parser.act
```diff
@@ -308,17 +310,48 @@
...
 		@c = lex_state->buf.a[2];
-		if ((unsigned char) @c > 127) {
-			@!;
+		/* from http://www.pcre.org/current/doc/html/pcre2pattern.html#SEC5
+		 * 
+		 *  The precise effect of \cx on ASCII characters is as follows: if x is a lower case letter, it is converted to
+		 *  upper case. Then bit 6 of the character (hex 40) is inverted. Thus \cA to \cZ become hex 01 to hex 1A (A is
+		 *  41, Z is 5A), but \c{ becomes hex 3B ({ is 7B), and \c; becomes hex 7B (; is 3B). If the code unit following
+		 *  \c has a value less than 32 or greater than 126, a compile-time error occurs. 
+		 */
```
katef, line 339: If this is specific to pcre, we might want to split `CONTROL` into two tokens, and provide other behaviour for other dialects which use it. Thanks for the comment here, it's clear enough that we can see to do that in the future if neccessary.

katef: Do you think they designed this syntax, or was it machine evolved?

katef (approved):

katef:
Brilliant work. Thank you again.

include/re/re.h
sfstewman: I actually like that it does.  People can write long, complex regexps, and comments along with the PCRE/Perl extended syntax makes them a lot more readable and maintainable.

PR #267, opened 2020-10-10
Fix bad comment handling

hvdijk:
This renames RE_BADCOMMENT to RE_EBADCOMMENT to fit the existing naming
convention, handles it in re_strerror, and adds a test where re would
previously hit the assert(!"unreached").


katef (approved):
Thanks!

PR #268, opened 2020-10-10
ast_new_from_fsm: free pool on error

hvdijk:
If ast_expr_new_from_fsm or ast_new fails, do not forget to free the pool.


katef (approved):
Thank you again!

PR #271, opened 2020-10-10
Another PCRE parser fix: ^ is literal in bracket expressions except directly after [

hvdijk:
There is no need to give ^ special treatment in the parser in bracket
expressions. If it appears immediately after [, it is handled by the
lexer by forming an $opengroupinv or $opengroupinvcb token. If it
appears elsewhere in a bracket expression, it is a literal.


katef:
I think this means we should also be able to get rid of `$invert` from the native and sql dialects in the same way, and then we can drop `TOK_INVERT` entirely.

hvdijk:
Native sure, but it's special in sql even if it does not appear directly after an opening bracket, `[a-z^d]` means "any character in `a-z` that's not `d`.

katef:
Oh yes, the binary operator form. Okay, so we'll always need the token around.

Would you mind updating both to handle the `[^` situation in the same way please? That at least keeps those parts consistent.

hvdijk:
Done. As discussed, this changes the way `-` is handled in the native dialect too to more closely match PCRE and to no longer make `[a-a]` an error.

katef (approved):

PR #272, opened 2020-10-11
Allow freed memory to be reused for AST nodes

hvdijk:
When ast_expr_free frees a node, it allows the memory used for that
node to be reused by a subsequent ast_make_expr_* call. During normal
parsing this will have no effect as memory will only be allocated, never
freed, but ast_new_from_fsm does mix allocations and deallocations.
ast_rewrite currently never allocates, only deallocates, but if future
improvements there do allocate, they will automatically benefit from
this as well.

Clearly marking already freed nodes this way also allows double frees to
be caught and reported. This commit contains a fix for a double free in
ast_rewrite.c's rewrite_concat function.

### DIFF
--- a/src/libre/ast_rewrite.c
+++ b/src/libre/ast_rewrite.c
@@ -24,7 +24,7 @@
 #endif
 
 static int
-rewrite(struct ast_expr *n, enum re_flags flags);
+rewrite(struct ast_expr_pool **poolp, struct ast_expr *n, enum re_flags flags);
 
 static int
 cmp(const void *_a, const void *_b)
@@ -36,11 +36,12 @@ cmp(const void *_a, const void *_b)
 }
 
 static void
-dtor(void *p)
+dtor(void *p, void *opaque)
 {
 	struct ast_expr *n = * (struct ast_expr **) p;
+	struct ast_expr_pool *pool = (struct ast_expr_pool *) opaque;
 
-	ast_expr_free(n);
+	ast_expr_free(pool, n);
 }
 
 /*
@@ -51,7 +52,8 @@ dtor(void *p)
 static size_t
 qunique(void *array, size_t elements, size_t width,
 	int (*cmp)(const void *, const void *),
-	void (*dtor)(void *))
+	void (*dtor)(void *, void *),
+	void *opaque)
 {
 	char *bytes = array;
 	size_t i, j;
@@ -68,7 +70,7 @@ qunique(void *array, size_t elements, size_t width,
 				memcpy(bytes + j * width, bytes + i * width, width);
 			}
 		} else {
-			dtor(bytes + i * width);
+			dtor(bytes + i * width, opaque);
 		}
 	}
 
@@ -115,7 +117,7 @@ compile_subexpr(struct ast_expr *e, enum re_flags flags)
 }
 
 static int
-rewrite_concat(struct ast_expr *n, enum re_flags flags)
+rewrite_concat(struct ast_expr_pool **poolp, struct ast_expr *n, enum re_flags flags)
 {
 	static const struct ast_expr zero;
 	size_t i;
@@ -125,7 +127,7 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
 	assert(n->flags == 0x0);
 
 	for (i = 0; i < n->u.concat.count; i++) {
-		if (!rewrite(n->u.concat.n[i], flags)) {
+		if (!rewrite(poolp, n->u.concat.n[i], flags)) {
 			return 0;
 		}
 	}
@@ -133,10 +135,6 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
 	/* a tombstone here means the entire concatenation is a tombstone */
 	for (i = 0; i < n->u.concat.count; i++) {
 		if (n->u.concat.n[i]->type == AST_EXPR_TOMBSTONE) {
-			for (i = 0; i < n->u.concat.count; i++) {
-				ast_expr_free(n->u.concat.n[i]);
-			}
-
 			goto tombstone;
 		}
 	}
@@ -144,7 +142,7 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
 	/* remove empty children; these have no semantic effect */
 	for (i = 0; i < n->u.concat.count; ) {
 		if (n->u.concat.n[i]->type == AST_EXPR_EMPTY) {
-			ast_expr_free(n->u.concat.n[i]);
+			ast_expr_free(*poolp, n->u.concat.n[i]);
 
 			if (i + 1 < n->u.concat.count) {
 				memmove(&n->u.concat.n[i], &n->u.concat.n[i + 1],
@@ -195,7 +193,7 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
 			n->u.concat.count += dead->u.concat.count;
 
 			dead->u.concat.count = 0;
-			ast_expr_free(dead);
+			ast_expr_free(*poolp, dead);
 
 			continue;
 		}
@@ -230,7 +228,7 @@ empty:
 tombstone:
 
 	for (i = 0; i < n->u.concat.count; i++) {
-		ast_expr_free(n->u.concat.n[i]);
+		ast_expr_free(*poolp, n->u.concat.n[i]);
 	}
 
 	free(n->u.concat.n);
@@ -242,7 +240,7 @@ tombstone:
 }
 
 static int
-rewrite_alt(struct ast_expr *n, enum re_flags flags)
+rewrite_alt(struct ast_expr_pool **poolp, struct ast_expr *n, enum re_flags flags)
 {
 	static const struct ast_expr zero;
 	size_t i;
@@ -252,7 +250,7 @@ rewrite_alt(struct ast_expr *n, enum re_flags flags)
 	assert(n->flags == 0x0);
 
 	for (i = 0; i < n->u.alt.count; i++) {
-		if (!rewrite(n->u.alt.n[i], flags)) {
+		if (!rewrite(poolp, n->u.alt.n[i], flags)) {
 			return 0;
 		}
 	}
@@ -299,7 +297,7 @@ rewrite_alt(struct ast_expr *n, enum re_flags flags)
 			n->u.alt.count += dead->u.alt.count;
 
 			dead->u.alt.count = 0;
-			ast_expr_free(dead);
+			ast_expr_free(*poolp, dead);
 
 			continue;
 		}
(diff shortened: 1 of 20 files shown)


sfstewman:
I think it's a good idea to add a free list!  I would have in the original implementation, but the normal AST pathways didn't benefit from it, so it seemed like extra code for no benefit.  If `ast_new_from_fsm` would benefit from it, we should add one!

That said, there are two things I'm not keen on in this implementation:
1. Chunks are immediately turned into a linked list at allocation.  I would prefer keeping the allocated chunks as-is and also adding a free list mechanism: on allocation, check the free list, if empty, proceed to the current chunk.  If the current chunk is full, allocate a new one.  Having a separate free list would add a (mostly predictable) branch in allocation, and avoid creating a large linked list when a chunk is allocated.
2. The `AST_EXPR_FREE` type.  AST nodes shouldn't be aware that they're in a freed state.  We can turn the `ast_expr_pool.pool` member into a union that holds this information and avoid exposing how we manage allocation and freeing the nodes to the users of the nodes.

Out of curiosity, how much memory is saved by reusing allocation in `ast_new_from_fsm`?

I'm not 100% sure that this plays well with ASAN poisoning.  Of course, I'm not 100% sure the current implementation plays well with ASAN poisoning, either.  ASAN poisoning needs more testing regardless of how the free list is implemented.

hvdijk:
Thanks for the comments!

> 1. Chunks are immediately turned into a linked list at allocation.  I would prefer keeping the allocated chunks as-is and also adding a free list mechanism: on allocation, check the free list, if empty, proceed to the current chunk.  If the current chunk is full, allocate a new one.  Having a separate free list would add a (mostly predictable) branch in allocation, and avoid creating a large linked list when a chunk is allocated.

Sure, that should not be a problem. There are two ways that this can be done: either by adding an extra field to `ast_expr_pool` (the `count` field that I had dropped in this PR), or by having two distinct "unallocated node" states, one which tells the allocator where the next node is via a pointer, another which tells the allocator that the next block in memory is the next node. The former is easiest to implement correctly, so unless there's a strong preference for the latter, I'll try to go with the former.

> 2. The `AST_EXPR_FREE` type.  AST nodes shouldn't be aware that they're in a freed state.  We can turn the `ast_expr_pool.pool` member into a union that holds this information and avoid exposing how we manage allocating and freeing the nodes to the users of the nodes.

I intentionally did it this way to make it possible to detect double frees, which I felt was a valuable debugging aid and caught a pre-existing bug already. Agreed that it'd be nice to find a way to not have it be part of the public API though. Will have a think about whether there is a clean way we can have both.

> Out of curiosity, how much memory is saved by reusing allocation in `ast_new_from_fsm`?

It depends massively on how big and complex the FSM is, but the worst case needs n² expressions for n states for each step the function takes (each step reducing the number of states by 1), so if expressions are never reused, that takes up the equivalent of n!² expressions in total. That is without taking into account the complexity of each individual expression.

> I'm not 100% sure that this plays well with ASAN poisoning. Of course, I'm not 100% sure the current implementation plays well with ASAN poisoning, either. ASAN poisoning needs more testing regardless of how the free list is implemented.

I did take this into account in this PR: in the code that I modified, I made sure to initialise the pool's unallocated memory before poisoning, and to only access it after unpoisoning.

sfstewman:
> > 1. Chunks are immediately turned into a linked list at allocation.  I would prefer keeping the allocated chunks as-is and also adding a free list mechanism: on allocation, check the free list, if empty, proceed to the current chunk.  If the current chunk is full, allocate a new one.  Having a separate free list would add a (mostly predictable) branch in allocation, and avoid creating a large linked list when a chunk is allocated.
> 
> Sure, that should not be a problem. There are two ways that this can be done: either by adding an extra field to `ast_expr_pool` (the `count` field that I had dropped in this PR), or by having two distinct "unallocated node" states, one which tells the allocator where the next node is via a pointer, another which tells the allocator that the next block in memory is the next node. The former is easiest to implement correctly, so unless there's a strong preference for the latter, I'll try to go with the former.

I think I prefer the former, as well.  It seems easier to implement and maintain.

> > 1. The `AST_EXPR_FREE` type.  AST nodes shouldn't be aware that they're in a freed state.  We can turn the `ast_expr_pool.pool` member into a union that holds this information and avoid exposing how we manage allocating and freeing the nodes to the users of the nodes.
> 
> I intentionally did it this way to make it possible to detect double frees, which I felt was a valuable debugging aid and caught a pre-existing bug already. Agreed that it'd be nice to find a way to not have it be part of the public API though. Will have a think about whether there is a clean way we can have both.

Ah.  That's a great point!  I think we can do both by doing something like this:

```
struct ast_expr_pool_item {
	union{
		struct ast_expr expr;
		struct {
			enum ast_expr_type type;
			struct ast_expr_pool_item *next;
		} freed;
	} u;

#if defined(ASAN)
	struct {
		void *ptr[8]; /* force 8-byte alignment on amd64 */
	} redzone;
#endif
};

#define AST_FREED_MARKER (~(enum ast_expr_type)0)
struct ast_expr_pool {
	struct ast_expr_pool_item pool[AST_EXPR_POOL_SIZE];

	struct ast_expr_pool *next;
	unsigned count;
};
```

We can then inspect the `type` field in both members of the `ast_expr_pool_item` union.  If the value is `AST_FREED_MARKER`, then we know that this is a double-free.

NB: This is explicitly allowed in C11 (6.5.2.3p6) and (I believe) C99.  I'm not sure about C90 (or C++).  @katef might want to weigh in on whether this approach is okay for libfsm.

> > Out of curiosity, how much memory is saved by reusing allocation in `ast_new_from_fsm`?
> 
> It depends massively on how big and complex the FSM is, but the worst case needs n² expressions for n states for each step the function takes (each step reducing the number of states by 1), so if expressions are never reused, that takes up the equivalent of n!² expressions in total. That is without taking into account the complexity of each individual expression.

Ouch!  Sounds like it would really benefit from this!

> > I'm not 100% sure that this plays well with ASAN poisoning. Of course, I'm not 100% sure the current implementation plays well with ASAN poisoning, either. ASAN poisoning needs more testing regardless of how the free list is implemented.
> 
> I did take this into account in this PR: in the code that I modified, I made sure to initialise the pool's unallocated memory before poisoning, and to only access it after unpoisoning.

I noticed that; I appreciate it!  That last comment was more a note to myself that this part of the code needs more testing.

hvdijk:
> We can then inspect the `type` field in both members of the `ast_expr_pool_item` union. If the value is `AST_FREED_MARKER`, then we know that this is a double-free.

I think to have the use of `AST_FREED_MARKER` be valid and stable, its value needs to be zero, and the node types need to start from 1. Otherwise, suppose `enum ast_expr_type`'s underlying type is `unsigned char`. `~(enum ast_expr_type)0` will be `-1` because of integral promotions, which is outside of `enum ast_expr_type`'s range, and because `enum ast_expr_type` is not an unsigned type (an enum type is an integer type per 6.2.5p17, but never a signed or unsigned integer type per p4 and p6), the reduction modulo `UCHAR_MAX+1` does not apply if the value is out of range, and the behaviour is implicitly undefined.

Even if we say that this should be allowed and compilers will treat it as allowed, consider what happens when libre grows to have exactly 256 node types: `AST_FREED_MARKER` will clash with an existing node type.

Aside from that, C++ (since you brought that up :)) has very different rules for out-of-range enum values: it allows out-of-range enum values that use only those bits also used by the defined values (so that if an enum has `A=1`, `B=2`, they can be used as flags: the type must also be able to represent `0` and `A|B`), but it does not allow out-of-range values beyond that.

All of those problems are avoided by renumbering.

> NB: This is explicitly allowed in C11 (6.5.2.3p6) and (I believe) C99. I'm not sure about C90 (or C++). @katef might want to weigh in on whether this approach is okay for libfsm.

The common initial sequence rule applies to C90 and C++ as well, as far as I know. However, as commonly implemented in compilers, it requires access to be done through the union type. If the access is not done through the union type -- and libre would require ugly code to ensure the access is done through the union type -- this does not work in practice, and it is unclear to me whether it is the intent of the standard to allow it.

hvdijk:
I have updated the PR to not expose `AST_EXPR_FREE`, thus ensuring that failing to handle it in a `switch` statement will not trigger compilation warnings. I have done this by starting with `enum ast_expr_type { AST_EXPR_EMPTY = 1`. This turned out to require a small reworking of the code because now *handling* it in a `switch` statement will trigger compilation warnings, hence the `default: if (n->type) assert(!"unreached");` rather than adding a case for `AST_EXPR_FREE`/`(enum ast_expr_type)0`.

I've also restored `ast_expr_pool`'s `count` member.

I did keep `struct ast_expr`'s `u.free` union member since I did not manage to find a way to move that into a container union, or rather, since I did not manage to find a clean way that is valid and does not violate assumptions made by mainstream compilers, unfortunately.

katef:
Thank you both for this. Do you have an example that illustrates the double free during AST rewriting please?

src/libre/ast_rewrite.c
```diff
@@ -125,26 +127,22 @@ rewrite_concat(struct ast_expr *n, enum re_flags flags)
...
 		}
 	}
 
 	/* a tombstone here means the entire concatenation is a tombstone */
 	for (i = 0; i < n->u.concat.count; i++) {
 		if (n->u.concat.n[i]->type == AST_EXPR_TOMBSTONE) {
-			for (i = 0; i < n->u.concat.count; i++) {
-				ast_expr_free(n->u.concat.n[i]);
-			}
-
```
katef, line 139: Is this the double free bug?

katef: I tried to reproduce this with something like `./build/bin/re -r pcre -pl ast 'abc(?:$x)+def'` thinking that `(?:$x)` would become a tombstone, but that doesn't seem to do it.

hvdijk: Yup, that's the double free bug. `(?:$x)` doesn't become a tombstone, that might be a useful optimization to add. `[^\x00-\xff]` and `[x^x]` (in the SQL dialect) do, so any concatenation with those will trigger it. This shows up in `tests/native/in31.re`, `tests/native/in34.re`, `tests/native/in36.re`, `tests/pcre-anchor/in52.re`, and `tests/sql/in12.re`.

katef: From @hvdijk:
> Ah and now I see why this didn't show up as an error before the pool: the double free was introduced by commit 9338812e, which is part of the PR to use the pool, which immediately makes it harmless :)

katef:
We talked about this offline. I'm not a huge fan of adding the field to the union in `struct ast_expr`. However the corresponding node `.type` isn't exposed outside of the allocation, and more importantly it's mutually exclusive with the other node types. Conceptually that could be done as a union outside of this struct, but then that'd need a tagged enum too, and that's just wasteful when we're trying to save memory already. So I'm happy with this arrangement, and I have no further suggestions.

Thank you! I'm glad for the allocation re-use, especially with more AST rewriting to come.

I'm still not a fan of needing the pool in the first place, but that's a separate concern.

PR #279, opened 2020-11-02
tests/union: Add test case exercising unioning non-0 start states.

silentbicycle:
This adds test FSM inputs where the start state is non-0. `fsm_union` handles this properly, but it seems like a good case to check

### DIFF
--- a/tests/union/in3a.fsm
+++ b/tests/union/in3a.fsm
@@ -1,0 +1,13 @@
+#
+# Copyright 2008-2017 Katherine Flavel
+#
+# See LICENCE for the full copyright terms.
+#
+
+0 -> 1 'a';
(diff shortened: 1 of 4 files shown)


tests/union/in3a.fsm
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright 2008-2017 Katherine Flavel
```
katef, line 2: Wrong author! (also for minimise.c now!)

silentbicycle: Test header copyright fixed in 146b4f8c.

What should I do for minimise.c? While I substantially rewrote it, it's been part of the project for a while.

katef (approved):

PR #280, opened 2020-11-04
Implementing DFA-based group capture matching, part 1

silentbicycle:
This is the first pass of work adding support for capture actions in the DFAs.

When multiple DFAs with captures are unioned and then determinised (but not minimised after), we can track info about captures for them. (Minimisation after is not supported because it combines states in a way that leads to false positives)

- Add `fsm_union_array` to the public interface. This supports unioning an array of fsms, internally tracking the fsms' start states and base capture offsets in the combined fsm.

- Update the interfaces related to merge, union, and concat, so the base/capture_base info is populated.

- The combined DFA will have some capture false positives -- it may set captures for other DFAs executed on the same path, but there's currently no association between end states and the original DFAs, so they cannot be filtered out. Work in a future PR should add this; for now, a flag in the `capture_string_set_invariants` and some other regression tests enables/disables checking for these false positives. Some of those tests have pending work to track this better once implemented, but for now the minimal case that triggers this is unioning `aaa` and `(a)a` and giving the input string "aaa" (`capture_string_set_regression_false_positive2`).

- Similarly, there is a temporarily ignored ambiguity in `tests/capture/capture4.c`: determinization loses the information that "abc" matches both `ab*c` AND `abc`. This is noted in the test, and will be worked on next, but this pass is focused on the capture action mechanism itself.

- Zero-width captures (such as `ab()cd()e`) are explicitly not supported. They lead to ambiguities when they're preceded by other greedy matches. Fortunately, PCRE also does not support them.

### DIFF
--- a/src/libfsm/internal.h
+++ b/src/libfsm/internal.h
@@ -40,6 +40,8 @@ struct state_array;
 
 #define FSM_ENDCOUNT_MAX ULONG_MAX
 
+#define FSM_CAPTURE_MAX INT_MAX
+
 struct fsm_edge {
 	fsm_state_t state; /* destination */
 	unsigned char symbol;
@@ -47,6 +49,10 @@ struct fsm_edge {
 
 struct fsm_state {
 	unsigned int end:1;
+
+	/* If 0, then this state has no need for checking
+	 * the fsm->capture_info struct. */
+	unsigned int has_capture_actions:1;
 
 	/* meaningful within one particular transformation only */
 	unsigned int visited:1;
@@ -67,6 +73,7 @@ struct fsm {
 	fsm_state_t start;
 	unsigned int hasstart:1;
 
+	struct fsm_capture_info *capture_info;
 	const struct fsm_options *opt;
 };
 
--- a/src/libfsm/union.c
+++ b/src/libfsm/union.c
@@ -6,6 +6,7 @@
 
 #include <assert.h>
 #include <stddef.h>
+#include <string.h>
 #include <limits.h>
 #include <errno.h>
 
@@ -16,6 +17,8 @@
 #include <fsm/options.h>
 
 #include "internal.h"
+
+#define LOG_UNION_ARRAY 0
 
 struct fsm *
 fsm_union(struct fsm *a, struct fsm *b,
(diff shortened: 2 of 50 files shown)


tests/capture/captest.c
```diff
@@ -0,0 +1,239 @@
...
+			    (void *)eo_src);
+		}
+		assert(eo_src->tag == CAPTEST_END_OPAQUE_TAG);
+		eo_dst->ends |= eo_src->ends;
+		if (log_level > 0) {
+			fprintf(stderr, "carryopaque: eo_src ends -> 0x%x\n",
+			    eo_dst->ends);
+		}
+
+		fsm_setopaque(src_fsm, src_set[i], NULL);
```
silentbicycle, line 225: After some further work I'm not sure this actually should be clearing the src state's opaque here. Removing this seems sufficient to fix the suppressed failure in `capture4.c`, because it's no longer clearing the opaque and the `abc` end state is now considered a match for both `abc` and `ab*c`.

I'm not 100% sure this is correct, because I don't fully understand the invariants behind carryopaque yet. Does that make sense?

src/libfsm/internal.h
```diff
@@ -48,6 +50,10 @@ struct fsm_edge {
 struct fsm_state {
 	unsigned int end:1;
 
+	/* If 0, then this state has no need for checking
+	 * the fsm->capture_info struct. */
+	unsigned int has_capture_actions:1;
+
```
katef, line 56: :+1:

src/libfsm/union.c
```diff
@@ -80,3 +89,52 @@ fsm_union(struct fsm *a, struct fsm *b)
...
+
+#if LOG_UNION_ARRAY
+	for (i = 0; i < fsm_count; i++) {
+		fprintf(stderr, "union_array: bases %u: %zu, %zu\n",
+		    i, bases[i].state, bases[i].capture);
+	}
+#endif
+
+	return res;
+}
```
katef, line 140: :+1:

katef (approved):
I don't think anything I can say here really feels adequate :)

PR #282, opened 2020-11-09
Comment some model state machines for FSM operations

katef:
An attempt to diagram some model state machines, to serve as context for both the code and for more precise comments. I'm not sure I like these. Maybe they help, maybe they're too vague to be useful, maybe they do serve some purpose.

I haven't commented the typical Thompson NFA construction in the regexp AST to NFA transformation, because that area is likely to change as we move towards introducing Aho-Corasick and then work towards constructing Glushkov NFA directly.

### DIFF
--- a/src/libfsm/concat.c
+++ b/src/libfsm/concat.c
@@ -67,10 +67,12 @@ fsm_concat(struct fsm *a, struct fsm *b,
 	fsm_setend(q, ea, 0);
 
 	/*
-	 * The canonical approach is to create epsilon transition(s) from the end
+	 * The textbook approach is to create epsilon transition(s) from the end
 	 * state of one FSM to the start state of the next FSM.
 	 *
-	 * TODO: diagram
+	 *     a: ⟶ ○ ┄ ◎
+	 *                     a b: ⟶ ○ ┄ ○ ⟶ ○ ┄ ◎
+	 *     b: ⟶ ○ ┄ ◎
 	 *
 	 * In this implementation, if multiple end states are present, they are
 	 * first collated together by epsilon transitions to a single end state.
--- a/src/libfsm/concat.c
+++ b/src/libfsm/concat.c
@@ -70,9 +70,9 @@ fsm_concat(struct fsm *a, struct fsm *b,
 	 * The textbook approach is to create epsilon transition(s) from the end
 	 * state of one FSM to the start state of the next FSM.
 	 *
-	 *     a: ⟶ ○ ┄ ◎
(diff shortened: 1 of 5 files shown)


src/libfsm/concat.c
```diff
@@ -67,10 +67,12 @@ fsm_concat(struct fsm *a, struct fsm *b,
...
 
 	/*
-	 * The canonical approach is to create epsilon transition(s) from the end
+	 * The textbook approach is to create epsilon transition(s) from the end
 	 * state of one FSM to the start state of the next FSM.
 	 *
-	 * TODO: diagram
+	 *     a: ⟶ ○ ┄ ◎
+	 *                     a b: ⟶ ○ ┄ ○ ⟶ ○ ┄ ◎
+	 *     b: ⟶ ○ ┄ ◎
```
ranweiler, line 75: I like these a lot, overall.

But, whenever a state _symbol_ changes (e.g. to/from final) in the FSM composition, it makes it much harder to instantly see what state came from where. At least in those cases, it could help to include node labels.

Something roughly like:
```
	      a: ⟶ ○ ┄ ◎
                     a₁  a₂
	                      a b: ⟶ ○ ┄ ○ ⟶ ○ ┄ ◎
                                       a₁  a₂   b₁   b₂
	      b: ⟶ ○ ┄ ◎
                     b₁   b₂
```
More padding than in the above would probably be helpful, but you get the idea.

katef: I wondered about bracketing them perhaps. Like er...
```
a b: ⟶ ○ ┄ ○ ⟶ ○ ┄ ◎
      ╰──┬──╯ ╰──┬──╯
         a       b
```

ranweiler: Ooh, I think that's an improvement, too.

My only concern is that the brackets might visually overpower the arrows, since they present longer (unbroken) lines than the arrow tails. But there's a similar concern with my sketch, where the state labels can be a bit noisy. Brackets would probably be ideal if there were a way to make them a slightly dimmed color.

What I do like about state labels is that my eye can bounce between two states and instantly see "a₂ was final before composition, but isn't any longer". In other words, the idea that a state in the new FSM is "the same" as one of the inputs is encoded in the preservation of the label, even if its properties (final or not) change (seen in the symbol).

That said, a bias is that I am personally _very_ used to such a convention from stuff like category theory, so it might just be clearer from familiarity. I guess it's worth looking at how these generalize to the other diagrams, too. Particularly `union`.

katef: Definitely replacing the `┄` with `○ ··· ◎`, that helps.

PR #284, opened 2020-11-18
stateset: make iterator's set argument const

silentbicycle:
The iterator's set argument should probably be const. It looks like this got missed.

This also has a commit to remove a line from test code, which will be removed in an upcoming branch, but here would need casting to modify the now-const pointer -- this would have been another clue that call was actually incorrect.


katef (approved):

katef:
Thanks!

PR #285, opened 2020-11-25
Fix some misc. lingering issues in the capture code

silentbicycle:
- Remove a flag that I thought was necessary to handle the logic properly for PCRE anchors. It wasn't actually necessary, but didn't removed, and still had a silly temporary name.
- Actually grow the capture action trail. Add a test that exercises this. This got missed, and would lead to an assert failure with any capture more than 8 characters long.
- Mark the seen edges during the capture action graph search. This was missing, and while it probably wouldn't affect correctness, it could cause exponential behavior in pathological cases (because of backtracking).
- Add a special case for states with self-edges, so they get appropriate `EXTEND` capture actions added to the DFA. This is necessary for cases like `a(b*)(c)`, where it would handle "ac" correctly (capturing the empty string and "c"), but "abbbc" would not extend the existing capture as it encountered each "b", erroneously leading to "a()bbb(c)" rather than "a(bbb)(c)". Add a test for this (capture5.c). I _think_ this is correct, but it has very subtle interactions with anything else. Only applying captures to DFAs means that we can avoid the inherently ambiguous cases I can think of, such as `a(b*)(b)`, and property testing isn't finding anything else, but it needs more thought. 
- Minor interface changes in the property test code, which is not built by default, and was broken by other interface changes (refining const-ness).

I found most of these by searching for `FIXME`; usually I do that before posting PRs, but these got missed.

Should `step_trail_iter_epsilons`'s " /* skipping this for now */" in `capture.c` be an assert? We currently only support DFAs (it just skips checking epsilon edges), but the interface doesn't really enforce that clearly (such as an `fsm_all(fsm, fsm_isdfa)` check upfront), it just ignores the epsilons.

### DIFF
--- a/src/libre/ast_analysis.c
+++ b/src/libre/ast_analysis.c
@@ -162,7 +162,7 @@ analysis_iter(struct analysis_env *env, struct ast_expr *n)
 }
 
 static int
-always_consumes_input(const struct ast_expr *n, int thud)
+always_consumes_input(const struct ast_expr *n)
 {
 	if (is_nullable(n)) {
 		return 0;
@@ -187,7 +187,7 @@ always_consumes_input(const struct ast_expr *n, int thud)
 		size_t i;
 
 		for (i = 0; i < n->u.concat.count; i++) {
-			if (always_consumes_input(n->u.concat.n[i], thud)) {
+			if (always_consumes_input(n->u.concat.n[i])) {
 				return 1;
 			}
 		}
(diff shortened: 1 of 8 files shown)


src/libre/ast_analysis.c
```diff
@@ -162,7 +162,7 @@ analysis_iter(struct analysis_env *env, struct ast_expr *n)
 }
 
 static int
-always_consumes_input(const struct ast_expr *n, int thud)
+always_consumes_input(const struct ast_expr *n)
```
katef, line 165: Very glad to see this go! I couldn't work out what it was! Thank you!

katef (approved):

PR #286, opened 2020-11-26
Code generation to awk

katef:
Here I'm adding code generation to awk. it's unfortunately gawk dialect, because of the switch.

I'm also using recursion to change state (rather than simulating a label as in the Rust codegen, #234). Perhaps that's more suitable for other languages; it's not a great idea in awk. Unlike the Rust codegen, I'm able to fall through between cases.

For a survey of TCO (or the lack of) in awk implementations, see https://blog.0branch.com/posts/2016-05-13-awk-tco.html

Here's how the generated code looks:
image

And the corresponding CFG of VM ops:
image

### DIFF
--- a/include/fsm/print.h
+++ b/include/fsm/print.h
@@ -16,6 +16,7 @@ struct fsm;
  * specified. The available formats are:
  *
  *  fsm_print_api    - C code which calls the fsm(3) API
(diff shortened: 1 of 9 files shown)


katef (commented):
Good job me

PR #287, opened 2020-11-28
CI with ASAN

katef:
I'm adding a run under ASAN, for leak detection.

This is probably annoying because we have a bunch of leaks in tests, and there doesn't seem to be a good way to ignore those. But I figure finding leaks in the libraries is important enough to make it worthwhile.

My intention is to merge this, fix up whatever needs fixing, and then rely on it for spotting leaks in new code.

Here I'm including both gcc and clang, just in case they find anything different. I've removed the `install` target (because we're not testing that), and just one bmake variant (because it's not the makefiles we're testing here).

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -1,0 +1,33 @@
+name: ASAN
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest ]
+        cc: [ clang, gcc ]
+        make: [ bmake ]
+    
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: recursive
+
+    - name: dependencies
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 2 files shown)


silentbicycle (commented):
Looks good. Is it possible to get debugging symbols included in the CI ASAN reports?

This will probably be neater after merging and rebasing on #288, which fixes several issues. There are also several error handling paths that leak memory (search for "/* TODO: free stuff */"), if those aren't caught here it's a sign we don't have tests exercising them.

katef:
Currently `lx -l test` leaks; I better fix that before merging this.

katef:
(or someone else is welcome to, too :D)

katef:
Maybe we can add running under valgrind, too. That'll need some makefile work though, I'll add that another time.

silentbicycle:
Debugging symbols seem to be working now.

katef:
In 548388aaf27fee965c0824e35bf20669e1b67fda I'm excluding lx from ASAN's leak detection; I just don't want to spend the time in that area right now. It's a short-lived CLI program, and its own datastructures aren't critical. I'm more interested in making sure we get things right for the libraries, first.

PR #288, opened 2020-11-29
Fix various leaks exposed by ASAN when running tests and reperf

sfstewman:
Running the usual tests and reperf with the leak detector in ASAN found a bunch of leaks:

- `fsm_vm_free()` was previously a stub that did nothing but leak VM data.  Now it actually frees the VM data (faf8abd)
- `reperf` itself had a number of leaks (b1d2c45, 963e0c5, f8e7d96)
- Capture data was not freed in `fsm_free()`, and now is (2480033 and fb4afee)  ping: @silentbicycle 
- The "opaque" data used by capture tests (see below).  ping: @silentbicycle 

This PR also includes an example `reperf` script (f13522b).

A note about the opaque data used by the capture tests:
As mentioned in the comment in [captest.c](https://github.com/katef/libfsm/commit/06dedbec7d72b6e069759d3d0af4a4a3d8be5cb9#diff-bd9ad04eaafdcbeab4448632ca77b3b835490c23f77fc4caf95ae065b8145aa2R16-R24):
```/* Pool allocator for opaque data to eliminates inadvertent memory leaks.
 *
 * XXX - this isn't great, but it's quite hard to keep track of opaque
 *       data through the various fsm operations.
 *
 *       The right solution is probably to have each fsm keep an opaque
 *       "destructor" that it calls when the states of an FSM are
 *       destroyed.
 */
```
The pool allocator is a hack here.  It would be much better to give libfsm a opaque destructor callback to inform the client whenever a piece of opaque data is no longer referenced.  However, that's far beyond the scope of this PR.

### DIFF
--- a/src/libfsm/vm.c
+++ b/src/libfsm/vm.c
@@ -127,7 +127,7 @@ fsm_vm_compile(const struct fsm *fsm)
 void
 fsm_vm_free(struct fsm_dfavm *vm)
 {
-	(void)vm;
+	dfavm_free_vm(vm);
 }
 
 static enum dfavm_state
--- a/src/libfsm/vm/vm.c
+++ b/src/libfsm/vm/vm.c
@@ -473,3 +473,18 @@ error:
 	return NULL;
 }
 
+void
+dfavm_free_vm(struct fsm_dfavm *vm)
+{
+	if (vm->version_major == DFAVM_VARENC_MAJOR && vm->version_minor == DFAVM_VARENC_MINOR) {
+		dfavm_v1_finalize(&vm->u.v1);
+	} else if (vm->version_major == DFAVM_FIXEDENC_MAJOR && vm->version_minor == DFAVM_FIXEDENC_MINOR) {
+		dfavm_v2_finalize(&vm->u.v2);
+	} else {
+		/* invalid VM version! */
+		/* XXX - should we do something? */
+	}
+
+	free(vm);
+}
+
--- a/src/libfsm/vm/vm.c
+++ b/src/libfsm/vm/vm.c
@@ -477,6 +477,10 @@ error:
 void
 dfavm_free_vm(struct fsm_dfavm *vm)
 {
+	if (vm == NULL) {
+		return;
+	}
+
 	if (vm->version_major == DFAVM_VARENC_MAJOR && vm->version_minor == DFAVM_VARENC_MINOR) {
 		dfavm_v1_finalize(&vm->u.v1);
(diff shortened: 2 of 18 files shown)


src/libfsm/vm.c
```diff
@@ -127,7 +127,7 @@ fsm_vm_compile(const struct fsm *fsm)
 void
 fsm_vm_free(struct fsm_dfavm *vm)
 {
-	(void)vm;
+	dfavm_free_vm(vm);
```
katef, line 130: well there's yer problem...

src/libfsm/vm/vm.c
```diff
@@ -473,3 +473,18 @@ dfavm_compile_vm(const struct dfavm_assembler_ir *a, struct fsm_vm_compile_opts
...
+void
+dfavm_free_vm(struct fsm_dfavm *vm)
+{
+	if (vm->version_major == DFAVM_VARENC_MAJOR && vm->version_minor == DFAVM_VARENC_MINOR) {
+		dfavm_v1_finalize(&vm->u.v1);
+	} else if (vm->version_major == DFAVM_FIXEDENC_MAJOR && vm->version_minor == DFAVM_FIXEDENC_MINOR) {
+		dfavm_v2_finalize(&vm->u.v2);
+	} else {
+		/* invalid VM version! */
+		/* XXX - should we do something? */
```
katef, line 485: Probably assert on it. It's not part of the calling contract to deal with invalid things passed here.

I would allow for `vm` to be NULL though, in the spirit of `free()`.

silentbicycle (commented):
The changes to my test code look good (though they'll probably be replaced shortly). The other changes look good to me.

tests/capture/captest.c
```diff
@@ -12,6 +13,70 @@
 	    __FILE__, __LINE__, MSG);		\
 	exit(EXIT_FAILURE)
 
+/* Pool allocator for opaque data to eliminates inadvertent memory leaks.
```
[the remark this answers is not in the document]
sfstewman: I generally agree that leaks in test code are often fine.  Unfortunately, the ASAN leak detector is a really blunt tool, so it's hard to tell it that specific leaks are fine, but others are not.

The problem I run into is that any leaks will cause the tests to fail, which makes using ASAN difficult during development unless you disable the leak detector.  Which makes detecting leaks harder.  I haven't hit on a happy compromise, so I've ended up spending considerable time plugging leaks in various places, many of which don't actually matter.

I look forward to the end ID mechanism!  That sound quite nice.

src/libfsm/vm/vm.c
sfstewman: I think I've fixed these in 01c3fa9 and 67f3978

katef (approved):

PR #289, opened 2020-11-29
Fixes various compiler warnings

sfstewman:
This fixes a few compiler warnings that gcc 9.3 identifies.

The biggest change is the `fgets` handling in `reperf`.  gcc issued a warning that the return value was being ignored.  I've replaced it with a loop that correctly handles the full range of `fgets` behavior.


katef (approved):
Thank you!

PR #291, opened 2020-12-01
Various improvements to reperf

sfstewman:
Adds:
- `-e` option (for echo).  This describes each line read and what it does.  Useful for debugging.
- `Q` script command to quit the script early.
- `-C` flag to disable regexp compile measurements (which can be quite time consuming)
- `-N <cycles>` option to override the `N` commands for each line in the script.


katef (approved):
Thank you!

PR #292, opened 2020-12-01
DFA match group part two, end IDs

silentbicycle:
Add a new set of operations to the API, for "end IDs": before combining multiple DFAs with `fsm_union`, make it possible to associate a numeric ID with each's end states, so that reaching a particular end state in the combined DFA can be associated with a match of one _or more_ of the original DFAs. This is needed to filter out false positives with group captures, because partially overlapping paths in the combined state machine may set captures that should only be valid when halting at specific end states. For example, with `(ab*)(c)` vs`(ab)(c)`, when there is `a`, exactly one `b`, then `c` the input matches both, but any other number of `b`s only matches the former, yet the capture for `(ab)` will still be set for "abbbbc".

There are several changes to preserve and convert this end metadata through transformations. This was previously done in a very ad hoc way through the `opaque` interface, but that would not compose with other uses of that interface, and since the use case for end IDs is better specified, the library itself can manage the dataflow better.

Some of this functionality is still under development -- the next in this series of PRs will add a standalone command line program which reads a list of regexes from a file, combines them into a single DFA, and then reads input and prints which regexes matched and the contents of any captures. Several things still need to be hooked up to support this in libre (such as associating `(` `)` in the AST with match groups), and integrating that will probably find some details that got missed.

### DIFF
--- a/include/fsm/fsm.h
+++ b/include/fsm/fsm.h
@@ -223,12 +223,16 @@ fsm_setendid(struct fsm *fsm, fsm_end_id_t id);
  * If id_buf has enough cells to store all the end IDs (according
  * to id_buf_count) then they are written into id_buf[] and
  * *ids_written is set to the number of IDs. The end IDs in the
- * buffer will be in ascending order (FIXME: not currently
- * enforced).
+ * buffer may appear in any order, but should not have duplicates.
  *
  * Returns 0 if there is not enough space in id_buf for the
  * end IDs, or 1 if zero or more end IDs were returned. */
-int
+enum fsm_getendids_res {
+	FSM_GETENDIDS_NOT_FOUND,
+	FSM_GETENDIDS_FOUND,
+	FSM_GETENDIDS_ERROR_INSUFFICIENT_SPACE = -1
+};
+enum fsm_getendids_res
 fsm_getendids(const struct fsm *fsm, fsm_state_t end_state,
     size_t id_buf_count, fsm_end_id_t *id_buf,
     size_t *ids_written);
--- a/include/fsm/fsm.h
+++ b/include/fsm/fsm.h
@@ -222,7 +222,9 @@ fsm_setendid(struct fsm *fsm, fsm_end_id_t id);
 /* Get the end IDs associated with an end state, if any.
  * If id_buf has enough cells to store all the end IDs (according
  * to id_buf_count) then they are written into id_buf[] and
- * *ids_written is set to the number of IDs.
+ * *ids_written is set to the number of IDs. The end IDs in the
+ * buffer will be in ascending order (FIXME: not currently
+ * enforced).
  *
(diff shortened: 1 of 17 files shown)


include/fsm/fsm.h
```diff
@@ -201,6 +206,41 @@ fsm_setopaque(struct fsm *fsm, fsm_state_t state, void *opaque);
...
+ * input when multiple fsms are combined.
+ *
+ * These will be preserved through the following operations:
+ * - determinise
+ * - union
+ * - concat
+ * - ...
+ * */
+int
+fsm_setendid(struct fsm *fsm, fsm_end_id_t id);
```
katef, line 220: What happens for other operations? Do they get cleared?

silentbicycle: I'm going to refine that comment in the next pass. If I can't support something I would rather clear it and explicitly note it there. We could also clear the capture actions if we detect that we're minimising after a union.

src/libfsm/endids_internal.h
```diff
@@ -0,0 +1,42 @@
...
+#define BUCKET_NO_STATE ((fsm_state_t)-1)
+#define DEF_BUCKET_COUNT 4
+#define DEF_BUCKET_ID_COUNT 16
+
+struct endid_info {
+	/* Add-only hash table, with a state ID and an associated
+	 * non-empty ordered array of unique end IDs. The state is the
+	 * key. Grows when the buckets are more than half full. */
+	unsigned bucket_count;
+	unsigned buckets_used;
```
katef, line 30: I suppose this is likely to get centralised under src/adt eventually. We have a couple of similar hash tables there already.

src/libfsm/glushkovise.c
```diff
@@ -318,3 +334,95 @@ remap_capture_action_cb(fsm_state_t state,
...
+
+	env->ids[env->count] = id;
+	env->count++;
+
+	return 1;
+}
+
+/* fsm_glushkovise can't use fsm_endid_carry directly, because the src
+ * and dst FSMs are the same -- that would lead to adding entries to a
+ * hash table, possibly causing it to resize, while iterating over it.
```
katef, line 377: Interesting situation.

katef (approved):

src/libfsm/endids_internal.h
silentbicycle: I figured as much, but didn't want to do it prematurely. A dynamic hash table of type `state_id -> 'a`, but some of them are add-only (and much simpler for it), and they're internally storing different sizes of values, so it could get a bit messy.

src/libfsm/glushkovise.c
silentbicycle: I suspect it was also the root cause of some mysterious use-after-frees I was seeing in the carryopaque callback. Since it was an odd use case and only in test code, I didn't investigate further.

PR #297, opened 2020-12-30
Determinise using interned state set

silentbicycle:
Add a wrapper around `state_set` construction that interns them, and use it in `fsm_determinise`.

While investigating ways to improve determinisation performance, I realized that it was spending the majority of its time constructing `state_set`s, checking if they existed, realizing they were redundant, and then freeing them. Moving the cache check into construction itself speeds this up considerably; each `state_set` is cached, and the wrapper keeps track of other derived `state_set`s that are created by adding a single state to an existing set. There is also a global cache, so if an equivalent set is incrementally built up in a different order, it will still converge on the same set, and get cached on the other interned state set. Then, we can use pointer equality to compare the `state_set`s in the transitive-closure-to-DFA-state mapping, rather than checking that every individual state in each set matches, which eliminates a lot of overhead in the mapping's bookkeeping.

This needs further testing and benchmarking, but casual benchmarking suggests that it's roughly twice as fast.

This looks like a gnarlier diff than it is; the overarching logic hasn't changed that much, but the specific data structure types have, so there are a lot of small changes all over the place. It uses a couple bespoke data structures, rather than the `adt` ones, because it needs two fairly simple add-only hash tables which are hashing an interned pointer (a unique ID) or a single state ID and don't need most of the adt hash table interfaces. (`siphash` is also pretty heavyweight for this use case, and dominates profiling.) The interned state set is added to adt/.

I've done some testing for memory leaks, but haven't thoroughly exercised the error handling yet; I was more interested in checking the tests and performance for a couple test data sets I've been using. All the existing tests pass.

The next step would be switching state_set from an ordered array to something with cheaper insertion. I have a prototype of this on another branch, but found that then instead of `state_set_add` dominating the profile, then it became state set comparison, and realized why `fsm_determinise` was spending nearly all of its time in `mapping_find`. Rather than using statehashset, I tried using a new implementation that hashed as a sum of `PHI32 * (state + 1)` for each state, because (unless I'm horribly mistaken) those will converged on the same hash in any order, and we can incrementally construct the hash -- in other words, it becomes cheap to check whether adding a state to an existing state_set would produce an existing, cached state_set without needing to actually construct it, which should speed up determinisation further (as will using a state set that is cheaper to insert into). Instead it would probably be better to integrate this idea into `statehashset.c` and use that instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ SUBDIR += tests/trim
 SUBDIR += tests/union
 SUBDIR += tests/set
 SUBDIR += tests/stateset
+SUBDIR += tests/internedstateset
 SUBDIR += tests/sql
 SUBDIR += tests/hashset
 SUBDIR += tests/queue
--- a/include/adt/internedstateset.h
+++ b/include/adt/internedstateset.h
@@ -30,8 +30,7 @@ interned_state_set_pool_alloc(const struct fsm_alloc *a);
 void
 interned_state_set_pool_free(struct interned_state_set_pool *pool);
 
-/* Get a handle to the empty state set.
- * Returns NULL on alloc failure. */
+/* Get a handle to the empty state set. Can not fail. */
 struct interned_state_set *
 interned_state_set_empty(struct interned_state_set_pool *pool);
 
--- a/include/adt/internedstateset.h
+++ b/include/adt/internedstateset.h
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2020 Scott Vokes
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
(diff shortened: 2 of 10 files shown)


katef:
I added new CI in #298; you may want to rebase on main or merge that, to get the new builds.

katef:
> The next step would be switching state_set from an ordered array to something with cheaper insertion.

I think the only use of `state_set_remove()` is when merging states; if we could account for that in some other way (perhaps by combining hashed entries), we could get rid of `state_set_remove()` and perhaps that allows for something that's cheap to insert without needing to remove items at all.

silentbicycle:
Rebased onto `main` with #298.

Issue #294, opened 2020-12-07
MacOS install fails on undefined symbol _fsm_capture_dump


sanderevers:
On MacOS Mojave with bmake version 20200902 from Homebrew.
`bmake -r install` ends in
```
[...]

PR #298, opened 2021-01-02
Rework CI and introduce coverage for both debug & release builds

katef:
This PR brings in various CI work that I experimented with in katef/kgt (which has a simpler build). In particular the yaml for the santizers is combined, and we now have coverage for both debug and release builds.

I haven't included CI for emscripten and packaging wasm here; for libfsm those depend on handling the libraries with support in kmkf. I'll do that separately because my goal here is just to get better test coverage for regular builds at the moment.

This should include coverage for the situation reported in #294.


katef:
Unsure why these checks show as pending:
image

As far as I can see (checking the combinations in yaml by eye), I think all are accounted for. Maybe I'll try merging this PR and see if that persists. I think perhaps it could be because the checks currently on `main` are now no longer present? I'm not sure at all.

katef:
with no way to debug this, I'm going to merge this branch and see what happens.

katef:
#299 shows those mystery checks are still required, so this needs resolving somehow. I suppose that means they must be coming from the yaml matrix stuff. Will debug on that branch.

katef:
I deleted all the .yml files and those expected checks still show 😬

katef:
aha! they were listed in the settings as required checks there. so they were old names. now I've unchecked those in the settings, the PR doesn't show them any more. mystery solved!

PR #299, opened 2021-01-02
[XXX: test for CI]

katef:
test for #298


katef:
all resolved, see #298 for discussion

Issue #300, opened 2021-01-22
vmc generates incorrect code for regexp with empty alt


katef:
```
; ./build/bin/re -pl c -r pcre '|'             

int
fsm_main(int (*fsm_getc)(void *opaque), void *opaque)
{
	int c;


	enum {
		S0
	} state;

	state = S0;

	while (c = fsm_getc(opaque), c != EOF) {
		switch (state) {
		case S0: /* start */
			return TOK_UNKNOWN;

		default:
			; /* unreached */
		}
	}

	/* end states */
	switch (state) {
	case S0: return 0x1; /* "|" */
	default: return -1; /* unexpected EOT */
	}
}

; ./build/bin/re -pl vmc -r pcre '|'

int
fsm_main(int (*fsm_getc)(void *opaque), void *opaque)
{
	int c;

	if (c = fsm_getc(opaque), c == EOF) return 0x1; /* "|" */
	return -1;
}

```

katef:
I think we construct the wrong VM opcodes:

![image](https://user-images.githubusercontent.com/1371085/105446825-c85dc300-5c27-11eb-8cd9-f9efc14770a1.png)

katef:
in `initial_translate_state()` for `IR_NONE` we construct:
```c
*opp = opasm_new_stop(a, VM_CMP_ALWAYS, 0, VM_END_FAIL, st);
```

I think this should accept when the `IR_NONE` node accepts.

I'm not sure if it's possible to construct an IR where the node does not accept, perhaps from an AST tombstone?

PR #302, opened 2021-01-25
Bugfix for VM-generated code from an accepting IR_NONE node

### LINKED ISSUE #301: Trailing comma in IR json
```
./build/bin/re -pl irjson -y tests/ir/in10.re
{
	"start": 0,
	"states": [
		{
			"end": true,
			"strategy": "none",
		}
	]
}
```

Probably a special case for `IR_NONE`

katef:
This PR addresses #300, where the codegen from VM opcodes was incorrect. Actually the codegen part was fine, but we constructed the wrong VM opcode by not taking into account whether an IR node accepts or not.

Note that the test here only tests which IR node is constructed for this situation; distressingly, we don't have coverage for executing the generated code. I hope for this to be covered through fuzzing instead.

While I'm here, I also fixed #301, where the json output had a syntactically incorrect trailing comma in some situations.


tests/ir/in10.re
```diff
@@ -0,0 +1 @@
+^$|^$
```
silentbicycle, line 1: I don't think this is actually testing the change properly -- running `build/bin/re -p -rpcre -lirjson '^$|^$'` on `main` still results in 
```
{
	"start": 0,
	"states": [
		{
			"end": true,
			"strategy": "none",
		}
	]
}
```
which is the same as the test below, give or take the trailing comma after `"strategy": "none"`.

Do we have example inputs `main` does not accept, but should, and the change fixes it? I thought that e.g. `/^abcde$|/` was one of them, but when I try to check that I get an assertion failure:
```
$ build/bin/re -p -rpcre '^abcde$|' bananas
re: src/re/main.c:383: void carryopaque(const struct fsm *, const fsm_state_t *, size_t, struct fsm *, fsm_state_t): Assertion `fsm_getopaque(src_fsm, src_set[i]) != NULL' failed.
Aborted (core dumped)
```
That may be an unrelated issue though.

silentbicycle (changes requested):
This needs an integration test, a regex and input that didn't match previously but does with the code change. The current test seems to pass without the code change, or fail for unrelated reasons (comparing via `diff` and a trailing comma).

tests/ir/in10.re
silentbicycle: The assert seems to fail both on this branch and `main`, and the simplest input that triggers it seems to be `build/bin/re -p -rpcre 'a|' a`. The `|` and `-p` are both necessary.

sfstewman:
> Note that the test here only tests which IR node is constructed for this situation; distressingly, we don't have coverage for executing the generated code. I hope for this to be covered through fuzzing instead.

If I understand what you mean correctly, the `retest` program should be able to handle this kind of test.  Both for `vm` interpreted code and `vmc` generated code.

The following `test_0.tst` input file fails on main and passes with `kate/vm-accepting-none`:
```# empty pcre ALT expression.  unanchored, so this should match anything
|
+a
+aaa
+

```

To run:
```# test interpreter:
./build/bin/retest test_0.tst

# test vmc generated code:
./build/bin/retest -l vmc test_0.tst
```

I've been meaning to integrate `retest` tests into the existing unit tests.  This seems like a reasonable place to start that.

katef:
yep, we really should have CI running retest suites. not on this PR though, that's a whole bunch of work

tests/ir/in10.re
katef: > I don't think this is actually testing the change properly -- running build/bin/re -p -rpcre -lirjson '^$|^$' on main still results in

that's right - we always did produce the IR_NONE node, and that isn't changing by this PR. I wanted to add it to the corpus just to be sure we do produce IR_NONE for that situation.

What's changing here is how IR_NONE gets converted to VM opcodes (we used to always construct VM_END_FAIL, and now that's done conditionally), and we don't have test coverage for any of the VM opcodes.

silentbicycle: I think the assert failure is happening because of clashing assumptions between glushkovise, which sets the state as an end (https://github.com/katef/libfsm/blob/main/src/libfsm/glushkovise.c#L144) but some of those states may not have previously been end states and haven't had their opaques copied over yet (https://github.com/katef/libfsm/blob/main/src/libfsm/glushkovise.c#L151-L152). The `carryopaque` hook for `re` asserts that their opaque is non-NULL (https://github.com/katef/libfsm/blob/main/src/re/main.c#L383), so when fsm_glushkovise uses the callback to copy them over, the assert fails. (If I comment out the assert, it leads to the result I'd expect, complaining about ambiguity.) It seems like determinising an NFA created by unioning FSMs whose start state also has an end state in their epsilon closure triggers the assert.

Given some other edge cases I ran into with `fsm_glushkovise` (which is mutating the FSM in place, and may temporarily break various implicit invariants for opaques), this sort of issue isn't that surprising, but it may be really tricky to fix. The end ID changes should obsolete the getopaques mechanism (and already accounts for `fsm_glushkovise`), so we can probably just address this by removing them soon?

silentbicycle: It also means the assert failure should be specific to the `re` CLI, unless other carryopaque hooks have the same assert.

katef:
I rebased here, to bring in the test/retest infrastructure from #303, thank you @sfstewman.

I've added regression tests for `|` there, these run using the default retest options (i.e. matching with VM execution, not code generation). I think I should provide a way for that to execute using codegen as well. Maybe we can just iterate over them all from the makefile.

Separately to this PR, I'm also bringing in runs for the pcre suite which run using every combination of retest execution options (for codegen, and vm interpretation).

sfstewman:
> I rebased here, to bring in the test/retest infrastructure from #303, thank you @sfstewman.
👍 

> I've added regression tests for `|` there, these run using the default retest options (i.e. matching with VM execution, not code generation). I think I should provide a way for that to execute using codegen as well. Maybe we can just iterate over them all from the makefile.

I'm not sure that we need more than the VM execution for routine unit tests.  Just choosing one is very fast and supplies an end-to-end check on the VM IR.

We'd certainly want more for CI tests.

> Separately to this PR, I'm also bringing in runs for the pcre suite which run using every combination of retest execution options (for codegen, and vm interpretation).
Big 👍 !

katef:
> I'm not sure that we need more than the VM execution for routine unit tests.

i don't want to rely on the pcre suite for covering all the different IR node situations, and therefore the different codegen handling (as well as whatever per-codegen logic we might have).

I think tests/retest is the ideal place for us adding knowledge of regressions (as I'm doing in this PR), and I think it's the only place we have where we could guarantee coverage for a given situation when executing matches. So I'd like to add that (I'll do it on this PR), I expect it'll just be iterating over `vmc`, `asm`,etc. Just like I'm doing in CI for the pcre suite.

katef:
Merging this now; I'm working on introducing the pcre suite (and fuzzing) separately. I'd rather not leave this hanging around in the meantime, because the work for that is introducing its own side effects.

PR #304, opened 2021-01-31
A couple of retest//clock_gettime-related bugfixes

katef:
retest had a use-after-`fclose()` found by efence, but not by ubsan, asan, etc. That's nice to see; I'm glad efence is here.

`clock_gettime` seems to be hidden in the presence of `_POSIX_C_SOURCE=2` for the macos machines github provides, so I removed it and all seems well.

I updated kmkf to de-duplicate `${DIR}` now we have several programs from the same source directory (retest and friends). That had caused some rules to get removed, due to the duplicate prefix.

These were all present in CI, and had caused CI to fail on main. But I'd missed these because CI doesn't run when merging a PR from a forked repo. I'm not sure how to provide for that, and still allow PRs from forks.

### DIFF
--- a/src/retest/main.c
+++ b/src/retest/main.c
@@ -7,8 +7,6 @@
 #if __linux__
 /* apparently you need this for Linux, and it breaks macOS */
 #  define _POSIX_C_SOURCE 200809L
-#else
-#  define _POSIX_C_SOURCE 2
 #endif
 
 #include <unistd.h>
--- a/src/retest/main.c
+++ b/src/retest/main.c
@@ -1064,14 +1064,14 @@ process_test_file(const char *fname, enum re_dialect default_dialect, enum imple
 		impl_ready = false;
 	}
 
-	fclose(f);
-
 	if (ferror(f)) {
 		fprintf(stderr, "line %d: error reading %s: %s\n", linenum, fname, strerror(errno));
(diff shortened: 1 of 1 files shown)


src/retest/main.c
```diff
@@ -272,6 +270,17 @@ xopen(const char *s)
...
 
+static void
+xclock_gettime(struct timespec *tp)
+{
+	assert(tp != NULL);
+
+	if (clock_gettime(CLOCK_MONOTONIC, tp) != 0) {
+		perror("clock_gettime");
+		exit(EXIT_FAILURE);
+	}
```
sfstewman, line 281: Do we need to `abort()` here?  If there's no timing info, that's okay.  retest isn't really about measuring performance.

I'm ambivalent.  This is certainly easier logic to follow.

src/retest/main.c
```diff
@@ -7,8 +7,6 @@
 #if __linux__
 /* apparently you need this for Linux, and it breaks macOS */
 #  define _POSIX_C_SOURCE 200809L
-#else
-#  define _POSIX_C_SOURCE 2
```
sfstewman, line 11: If it interferes with macos, we should probably have an `#elif !defined(__APPLE__)`.

Unless we don't need it for any other arch.

PR #306, opened 2021-02-01
Add pcre test suite for CI

katef:
This isn't quite ready yet; I'm making the PR so I can keep track of what I'm doing.

The pcre suite is under a licence that isn't suitable for including in this repo directly. Here I'm having CI download the source, and that way we don't commit those tests. This is a CI action though, so you can't run this from the build system using make. Perhaps I should move the guts to the makefiles instead, to allow for that.

### DIFF
--- a/.github/workflows/san.yml
+++ b/.github/workflows/san.yml
@@ -78,7 +78,32 @@ jobs:
             < ${BUILD}/${PCRE_VER}/testdata/$i \
             > ${BUILD}/test/retest/$i.tst
         done
+        # things which will probably never be relevant for us:
+        rm ${BUILD}/test/retest/testinput2.tst # pcre options
+        rm ${BUILD}/test/retest/testinput3.tst # locale
+        rm ${BUILD}/test/retest/testinput5.tst # pcre options
+        rm ${BUILD}/test/retest/testinput8.tst # pcre bytecode
+        rm ${BUILD}/test/retest/testinput9.tst # character encodings
+        rm ${BUILD}/test/retest/testinput10.tst # character encodings
+        rm ${BUILD}/test/retest/testinput11.tst # character encodings
+        rm ${BUILD}/test/retest/testinput12.tst # character encodings
+        rm ${BUILD}/test/retest/testinput13.tst # character encodings
+        rm ${BUILD}/test/retest/testinput15.tst # backtracking limits
+        rm ${BUILD}/test/retest/testinput16.tst # JIT options
+        rm ${BUILD}/test/retest/testinput17.tst # JIT options
+        rm ${BUILD}/test/retest/testinput20.tst # serialisation
+        rm ${BUILD}/test/retest/testinput21.tst # \C
+        rm ${BUILD}/test/retest/testinput22.tst # \C
+        rm ${BUILD}/test/retest/testinput23.tst # \C
         rm ${BUILD}/test/retest/testinputEBC.tst # EBCDIC
+        # things we should eventually support, but don't yet:
+        rm ${BUILD}/test/retest/testinput4.tst # we don't provide UTF8 yet
+        rm ${BUILD}/test/retest/testinput7.tst # we don't provide UTF8 yet
+        rm ${BUILD}/test/retest/testinput14.tst # we don't provide UTF8 yet
+        rm ${BUILD}/test/retest/testinput18.tst # POSIX dialect
+        rm ${BUILD}/test/retest/testinput19.tst # POSIX dialect
+        rm ${BUILD}/test/retest/testinput24.tst # POSIX dialect
+        rm ${BUILD}/test/retest/testinput25.tst # POSIX dialect
 
     - name: run pcre test suite (vm v1)
       env:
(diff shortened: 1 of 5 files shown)


katef:
Comments from @sfstewman on which files to include (thank you!):

### Thoughts on the various PCRE2 `testinput` files

Based on the above descriptions, here are my thoughts:

1. `testinput1` is the basic test suite, and it pretty much works.  Ten tests fail; five of those
   are caused by backreferences that are harder to identify and five are caused by end-of-line
   behavior that libfsm does not yet support.
   
   Definitely include this file in the CI tests.

2. `testinput2` tests error diagnostics and various PCRE2 options for PCRE2 without UTF.  It has 
   some regexps that should fail, and don't necessarily.  To fully support this, we need to modify
   cvtpcre to simultaneously parse the `testoutput2` file and detect what should succeed and fail.
   
   We need to do that to fully test capture groups at some point, but it's kind of an undertaking.
   Also, many of the tests in `testinput2` test specific API features that PCRE2 supports.  I haven't
   thought carefully about these features, but many seem specific to a backtracking NFA engine, and
   we may never want to support.
   
   I would suggest not including this file in the CI tests.

3. `testinput3` switches the locale to `fr_FR`, `french`, or `fr` and tests some local-specific 
   features.  I'm not sure if we'd want to support this, but if we do, it will require some careful 
   thought.
   
   I would suggest not including this file in the CI tests at the moment.
   
4. `testinput4` is a set of tests for UTF support.  This is worth including.

   I would include this file and we'll eventually want to make sure that we pass every relevant test here.

5. `testinput5` is like `testinput2` but for UTF support.

   Similar to `testinput2`, I would suggest not including this file in the CI tests.

6. `testinput6` checks PCRE2's DFA matching without UTF support.  This is probably worth including, 
   but may be redundant with `testinput1`.
   
   I would include this file.

7. `testinput7` checks PCRE2's DFA matching with UTF support.  This is probably worth including, but
   may be redundant with `testinput4`.
   
   I would include this file.

8. `testinput8` looks like it's testing PCRE bytecode and memory allocation.  `testoutput8-*`
   seems to have the dumps for various encodings.
   
   Something like this would probably be valuable for the dfavm v1/v2 encodings, but these tests are
   very specific to PCRE2 internals, and I don't think they're valuable for us.
   
   I would suggest not including this file.

9. `testinput9` throught `testinput12` are tests for various widths of character input, with and without
   UTF and Unicode support.
   
   I'm not sure if these are relevant/useful for us.  The 8-bit library _might_ be.  But the 16-bit and
   32-bit versions are only relevant if we plan to support UTF-16 / UCS-16 and UTF-32 / UCS-32 inputs
   at some point instead of UTF-8.
   
   `testinput9` and `testinput10` are the 8-bit versions.  `testinput9` is for non-Unicode input and has a
   series of tests that should fail due to "too big char" errors.  This would require parsing `testoutput9`, 
   similar to what's required in `testinput2`.

   I would suggest not including any of these files for now.
   
   When cvtpcre can detect "expected error" conditions, we may want to consider `testinput9` and 
   `testinput10`, but possibly not.

10. `testinput13` tests PCRE2's DFA mode for handling chars > 255 in 16-bit and 32-bit non-UTF.  This does
    not seem relevant to us at the moment.
    
    I would suggest not including this file.

11. `testinput14` tests special UTF/UCP features of DFA matching.  "The output is different for
    different widths."
    
    I would suggest not including this file.

12. `testinput15` tests "match-limiting features."  This does not seem relevant to us.

    I would suggest not including this file.

13. `testinput16` and `testinput17` test the PCRE2 JIT.  None of the tests seem relevant.  They're
    Meant to examine the JIT state/output for different patterns and JIT options.
    
    I would suggest not including these files.

14. `testinput18` and `testinput19` test the POSIX interface with and without UTF support.  I think
    this means the API for POSIX-compatible regexps.
    
    I'm not sure if we support a POSIX dialect right now.  If we do, this might be useful, but we
    may also want to develop our own test suite for it.
    
    I would suggest not including these files.

15. `testinput20` tests "serialization/deserialization and code copy functions."  This doesn't seem
    relevant to libfsm at all.
    
    I would suggest not including this file.

16. `testinput21` and `testinput22` test `\C` for non-Unicode and Unicode situations.  From the PCRE spec: 
    "Outside a character class, the escape sequence `\C` matches any one code unit, whether or not a UTF 
    mode is set."
    
    This seems relevant, although we may not implement `\C` at the moment.
    
    I would suggest including these files (or at least `testinput22`) and eventually implementing `\C`.

17. `testinput23` tests PCRE2 behavior when `\C` has been disabled at compile time.  It basically tests
    for errors.

    I would suggest not including this file.

18. `testinput24` and `testinput25` test the "auxiliary pattern conversion features of the PCRE2 library"
    with and without Unicode/UTF support, respectively.
    
    Apparently PCRE2 can convert regexps to/from POSIX, glob, and maybe a few other dialects.  These files
    test this.
    
    This seems potentially relevant to libfsm, but ... we would need to modify `cvtpcre` to support this,
    similar to `testinput2`.

    I would suggest not including this file right now, and potentially including it later.

19. `testinputEBC` tests EBCDIC support.

    EBCDIC can diaf.
    
    I would suggest not including this file **EVER**.

katef:
oh my god

all tests passing

PR #309, opened 2021-02-03
Grouped edge_set ADT

silentbicycle:
Add grouped implementation of edge_set ADT.

This uses a bitset for the labels, since it's very common to have several labels that lead to the same state (in particular, "everything else goes here" edges) that could be beneficially combined, but it's comparatively rare to have the same label lead to lots of distinct states.

This currently leaves the existing implementation behind an `#if`, so that we can switch back and forth for benchmarking. Memory usage is likely better for large DFAs, but the callers will need to be adapted to truly benefit from these changes -- The iterator yields individual <label, state> pairs from each group to maintain compatibility with the existing implementation, but ideally some operations (especially `fsm_determinise`) should switch to processing groups of labels with the same destination together, because it will massively improve performance with large FSMs.

All tests are passing, including the edge_set properly tests, which were invaluable for testing.

Note: This adds various interface changes driven by edge_set operations that now alloc, and can therefore now fail.

These operations need an allocator handle for the edge_set label bit-set implementation. These changes should have little impact, because they're bubbling up an error for allocation failure, which is already largely non-recoverable, and fsm_removestate appears to only be called in fsm_mergestates, which is unused (and, per discussion, likely to be removed outside this PR).

### DIFF
--- a/include/adt/edgeset.h
+++ b/include/adt/edgeset.h
@@ -80,8 +80,8 @@ edge_set_remove(struct edge_set **set, unsigned char symbol);
 void
 edge_set_remove_state(struct edge_set **set, fsm_state_t state);
 
-void
-edge_set_compact(struct edge_set **set,
+int
+edge_set_compact(struct edge_set **set, const struct fsm_alloc *alloc,
     fsm_state_remap_fun *remap, const void *opaque);
 
 void
@@ -93,8 +93,9 @@ edge_set_next(struct edge_iter *it, struct fsm_edge *e);
 void
 edge_set_rebase(struct edge_set **setp, fsm_state_t base);
 
-void
-edge_set_replace_state(struct edge_set **setp, fsm_state_t old, fsm_state_t new);
+int
+edge_set_replace_state(struct edge_set **setp,
+    const struct fsm_alloc *alloc, fsm_state_t old, fsm_state_t new);
 
 int
 edge_set_empty(const struct edge_set *s);
--- a/include/adt/edgeset.h
+++ b/include/adt/edgeset.h
@@ -17,7 +17,7 @@ struct edge_set;
 struct state_set;
 
 struct edge_iter {
-	size_t i;
+	size_t i, j;
 	const struct edge_set *set;
 };
 
@@ -68,6 +68,8 @@ edge_set_transition(const struct edge_set *set, unsigned char symbol,
 size_t
 edge_set_count(const struct edge_set *set);
 
+/* Note: this should actually union src into *dst, if *dst already
+ * exists, not replace it. */
 int
 edge_set_copy(struct edge_set **dst, const struct fsm_alloc *alloc,
 	const struct edge_set *src);
--- a/include/fsm/fsm.h
+++ b/include/fsm/fsm.h
@@ -114,7 +114,7 @@ fsm_addstate_bulk(struct fsm *fsm, size_t n);
  /*
  * Remove a state. Any edges transitioning to this state are also removed.
  */
-void
+int
 fsm_removestate(struct fsm *fsm, fsm_state_t state);
 
 /* Use the state passed in via opaque to determine whether the state[id]
--- a/src/adt/edgeset.c
+++ b/src/adt/edgeset.c
@@ -9,6 +9,16 @@
 #include <string.h>
 #include <stdio.h>
 
+/* If non-zero, use an experimental edge_set implementation
+ * that collects edges leading to the same set in a group
+ * and uses bitflags for individual labels. While there should
+ * be some space savings vs. storing every <label, state> pair
+ * individually, the bigger motivation is to do processing for
+ * some operations (especially fsm_determinise) by group rather
+ * than for every individual pair. */
+#define USE_EDGE_BITSET 1
+
+#if !USE_EDGE_BITSET
 #include "libfsm/internal.h" /* XXX: for allocating struct fsm_edge, and the edges array */
 
 #include <print/esc.h>
@@ -18,6 +28,7 @@
 #include <adt/set.h>
 #include <adt/stateset.h>
 #include <adt/edgeset.h>
+
 
 /* This is a simple linear-probing hash table, keyed by the edge symbol.
  * Since many edge sets only contain a single item, there is a special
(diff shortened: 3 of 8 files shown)


include/adt/edgeset.h
```diff
@@ -17,7 +17,7 @@ struct edge_set;
 struct state_set;
 
 struct edge_iter {
-	size_t i;
+	size_t i, j;
```
silentbicycle, line 20: These names are vague because the struct is currently shared between the two implementations, which use the counters differently. The iterator isn't opaque because most of the callers stack-allocate it.

include/adt/edgeset.h
```diff
@@ -68,6 +68,8 @@ edge_set_transition(const struct edge_set *set, unsigned char symbol,
 size_t
 edge_set_count(const struct edge_set *set);
 
+/* Note: this should actually union src into *dst, if *dst already
```
silentbicycle, line 71: I was tripped up by this, twice (albeit months apart), because I don't think of this as typical 'copy' interface behavior.

include/adt/edgeset.h
```diff
@@ -78,8 +80,8 @@ edge_set_remove(struct edge_set **set, unsigned char symbol);
 void
 edge_set_remove_state(struct edge_set **set, fsm_state_t state);
 
-void
-edge_set_compact(struct edge_set **set,
+int
```
silentbicycle, line 83: This and `edge_set_replace_state` can now fail, but only due to allocation failure.

include/fsm/fsm.h
```diff
@@ -114,7 +114,7 @@ fsm_addstate_bulk(struct fsm *fsm, size_t n);
  /*
  * Remove a state. Any edges transitioning to this state are also removed.
  */
-void
+int
 fsm_removestate(struct fsm *fsm, fsm_state_t state);
```
silentbicycle, line 118: As discussed (on slack), the only current caller for `fsm_removestate` is `fsm_mergestates`, which does not appear to be used. We could potentially remove both from the public interface. `fsm_removestate` doesn't really fit well with the new continuous array of states implementation, we tend to compact on copy instead.

src/libfsm/consolidate.c
```diff
@@ -273,8 +273,8 @@ consolidate_end_ids_cb(fsm_state_t state, const fsm_end_id_t id,
 	assert(env->tag == 'C');
 
 #if LOG_CONSOLIDATE_ENDIDS > 1
-	fprintf(stderr, "consolidate_end_ids_cb: state %u, count %lu, IDs:",
-	    state, count);
+	fprintf(stderr, "consolidate_end_ids_cb: state %u, count %lu, ID %d:",
+	    state, id);
```
silentbicycle, line 277: Updating stale logging code, which no longer compiled when the log level was increased.

src/adt/edgeset.c
```diff
@@ -871,3 +887,997 @@ edge_set_ordered_iter_next(struct edge_ordered_iter *eoi, struct fsm_edge *e)
...
+	return 1;
+}
+
+static void
+dump_edge_set(const struct edge_set *set)
+{
+	const struct edge_group *eg;
+	size_t i;
+#if LOG_BITSET < 2
+	return;
```
silentbicycle, line 983: Do you have a preferred naming convention for this kind of thing? It logs the struct, but only if the log level is sufficiently high. `maybe_dump_edge_set`? `dump_edge_set_if_logging`?

src/adt/edgeset.c
```diff
@@ -871,3 +887,997 @@ edge_set_ordered_iter_next(struct edge_ordered_iter *eoi, struct fsm_edge *e)
...
+	 * it just exits at the first label found. */
+	if (set == NULL) {
+		return 0;
+	}
+
+	for (i = 0; i < set->count; i++) {
+		size_t w_i;
+		const struct edge_group *eg = &set->groups[i];
+		for (w_i = 0; w_i < 4; w_i++) {
+			/* TODO: res += popcount64(w) */
```
silentbicycle, line 1263: We could use a compiler intrinsic/builtin for 64-bit popcount here instead.

src/adt/edgeset.c
```diff
@@ -871,3 +887,997 @@ edge_set_ordered_iter_next(struct edge_ordered_iter *eoi, struct fsm_edge *e)
...
+	    (void *)set);
+#endif
+
+	if (edge_set_empty(set)) {
+		return 1;
+	}
+
+	assert(set->count > 0);
+
+	info = f_malloc(alloc, set->count * sizeof(info[0]));
```
silentbicycle, line 1536: I feel like there should be a way to do this that doesn't need 3 words per group, though in practice we're probably coming out way ahead by collapsing together lots of different labels leading to the same state.

My main problem with this implementation is that it still does as much work when the edge set's groups are already unique and sorted, compared to something that only rearranged/compacted in place as necessary, but my other approaches got tricky.

src/adt/edgeset.c
```diff
@@ -871,3 +887,997 @@ edge_set_ordered_iter_next(struct edge_ordered_iter *eoi, struct fsm_edge *e)
...
+
+int
+edge_set_empty(const struct edge_set *s)
+{
+	size_t i;
+	if (s == NULL || s->count == 0) {
+		return 1;
+	}
+
+	for (i = 0; i < s->count; i++) {
```
silentbicycle, line 1808: This needs to account for groups whose labels have all been removed, since currently we do not prune empty groups. It's not likely to be a common situation in practice, so the performance hit is probably okay. Property testing found a couple cases where that decision led to internal invariants breaking, all have been fixed.

src/adt/edgeset.c
katef: No preference from me. "dump" sounds fine (I tend to use that for datastructures).

src/adt/edgeset.c
katef: Nice.

src/adt/edgeset.c
katef: Do you know when that can happen?

katef (approved):
This is fantastic work, thank you so much

src/adt/edgeset.c
silentbicycle: You could ask theft by commenting out that part, adding `return 0;`, and running `build/theft/theft -n adt_edge_set`, it should produce at least two different kinds of failures in well under a second. :)

PR #311, opened 2021-02-03
Fixes off by one error in VM IR assembly logic

sfstewman:
The VM IR (used by the v1/v2 interpreters, vmc, etc.) has an off-by-one error when translating the IR_DOMINANT nodes.  Specifically, it ignores `\xff` characters.

This is picked up by the test case `/^\xff/` and the matching text `\xff`.


katef (approved):

sfstewman:
Moved test file to avoid conflicts with another PR.

PR #314, opened 2021-02-05
Fix typo in PCRE lexer.lx

hvdijk:
The unsupported "call subroutine by name" functionality uses (?P>name),
not (P>name), and this prevented capture groups from starting with P.


katef (approved):
oh! thank you!

PR #318, opened 2021-02-26
Update kmkf to conditionally elide rules for generating targets when not invoking `gen::` explicitly

katef:
Making this draft PR so I don't forget. I'm intending to fix the pcregrep CI error on main first, then rebase this (and write a proper description for it)

depreciating (and later removing) `.MADE`:
https://github.com/openbsd/src/commit/6c0d9dcd87ae1329dfcfced82fe671fc2514315b

### DIFF
--- a/src/libre/class/Makefile
+++ b/src/libre/class/Makefile
@@ -121,6 +121,7 @@ SRC += src/libre/class/${class}.c
 
 .endfor
 
+.if make(gen)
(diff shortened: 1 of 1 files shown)


katef:
Rebased on main; kmkf was since already updated to a commit which included the update for `gen::` here. So technically the "update kmkf" part of this PR was already done.

PR #319, opened 2021-03-01
Remove opaques interface (`fsm_setendopaques`, carryopaque, etc.), use end IDs interface instead

silentbicycle:
The opaques interface, particularly the `carryopaque` callback, exposes internal dataflow that we can handle properly internally if specific to tracking state end IDs (and metadata associated with them) through various operations and transformations. Maintaining the carryopaque interface has been an ongoing source of bugs (it's hard to use correctly, and there are `void *` casts surrounding it), and it won't be usable with upcoming changes to process groups of edges all at once.

Remove all those functions, remove the `carryopaque` function pointer from `struct fsm_options`, and change the `leaf` and `endleaf` callbacks' `state_opaque` parameter to a struct passing a set of end IDs. This is an interface change.

Update some tests to use end IDs that were still using opaques, or were using both.

Removing opaques and changing the `leaf`/`endleaf`'s `state_opaque` fixed a possible crash in lx, a memory safety issue that had been obscured by the void pointer casts.


silentbicycle:
There are three tests failing for me locally:

```
build/tests/pcre-pcregrep/res11:FAIL
build/tests/pcre-pcregrep/res12:FAIL
build/tests/pcre-pcregrep/res44:FAIL
```

but they also fail on `main`.

include/fsm/fsm.h
```diff
@@ -32,6 +32,13 @@ typedef unsigned int fsm_state_t;
  * their end states. These can be used to determine which of the
  * original FSM(s) matched when executing a combined FSM. */
 typedef unsigned int fsm_end_id_t;
+#define FSM_END_ID_NONE ((unsigned int)-1)
+
+/* struct used to return a collection of end IDs. */
+struct fsm_end_ids {
+	unsigned count;
+	fsm_end_id_t ids[1];
+};
```
katef, line 41: Did you introduce this because the set/adt stuff isn't exposed to users? That's perfect, thanks

include/fsm/fsm.h
```diff
@@ -32,6 +32,13 @@ typedef unsigned int fsm_state_t;
  * their end states. These can be used to determine which of the
  * original FSM(s) matched when executing a combined FSM. */
 typedef unsigned int fsm_end_id_t;
+#define FSM_END_ID_NONE ((unsigned int)-1)
```
katef, line 35: Is this used? I don't see it elsewhere in this diff

silentbicycle: Good catch. I think when I switched from a "get the ID" to "get the set of IDs" interface this was no longer used, but I forgot to remove it.

Fixed in 7bfffb0f.

include/fsm/fsm.h
silentbicycle: Right, this represents opaque numeric IDs associated with endpoints, but the internals know how to update them across operations. The caller can focus on what metadata is associated with them, but no longer needs a `carryopaque`-style interface that is coupled to the implementation details.

sfstewman:
Very exciting, @silentbicycle !  @katef asked me to leave a note here to remind us to add ID support to the dfavm IR, vm engines, and code generated from the dfavm IR.

silentbicycle:
I didn't look into it in detail, but as part of changing over current users of opaques to end IDs, I may have already done the first step (https://github.com/katef/libfsm/pull/319/files#diff-25aaa5dcaa6354d9b1b64ec5123186b04ccda87289904aa1e2163acb44d2d333R474) for the IR. As far as I know the only thing using that is some of the print code though.

silentbicycle:
Force-pushing with #320 merged, which appears to fix the local test failures that now started appearing in CI because whatever mysterious PCRE JIT heuristic started triggering there as well.

katef (approved):

katef:
With the ubsan and other various unrelated issues (from the build machines updating) now fixes on main, I'm going to merge this PR rather than having it rebased first. Because I'd prefer this one as a merge commit anyway.

Thank you, this is excellent work. Undoing my questionable choices from an awfully long time ago!

PR #320, opened 2021-03-01
tests/pcre/Makefile: call pcregrep with `--no-jit`. pcregrep bug?

silentbicycle:
The following tests started failing for me, on `main`:

 - build/tests/pcre-pcregrep/res11:FAIL
 - build/tests/pcre-pcregrep/res12:FAIL
 - build/tests/pcre-pcregrep/res44:FAIL

Sometimes they fail in CI, sometimes they don't. It seems to be related to `pcregrep`'s `-M` flag: removing the flag causes those
tests to pass, and the failure appears to have to do with multi-line mode.

@sfstewman found that pcregrep's `-M` option behaves differently with and without `--no-jit`. Confirmed locally, but I have not
investigated further. If it's a bug in heuristics related to JIT-ing, that would explain why it comes and goes.


katef (approved):
this is absolutely ridiculous. how frustrating. thank you both.

katef:
UBSan fails in the new edge set code; I presume that's unrelated. I'm going to merge this PR and we can handle that separately.

PR #321, opened 2021-03-02
Address several new warnings

silentbicycle:
Some of these are new in clang-9, some are from UBSan.

There are some type inconsistencies in our interfaces, UBSan is (understandably) complaining about us going back and forth between signed and unsigned char types implicitly with function parameters. I added casts, but should we just plan on using unsigned char throughout?

UBSan is also complaining at length about unsigned integer overflows in hash functions (particularly siphash), but the overflowing is intentional and (since it's unsigned) should still be defined behavior. Just adding casts is not satisfying UBSan, and the non-fixed-width numeric types will make addressing this a pain.

### DIFF
--- a/src/adt/internedstateset.c
+++ b/src/adt/internedstateset.c
@@ -116,7 +116,7 @@ interned_state_set_pool_alloc(const struct fsm_alloc *a)
 	{
 		const unsigned long mask = CACHE_BUCKET_DEF_CEIL - 1;
 		struct cache_bucket *b = &res->cache.buckets[empty_i->hash & mask];
-		assert(b->iss = NO_ISS);
+		assert(b->iss == NO_ISS);
 		b->hash = empty_i->hash;
 		b->iss = empty_i;
 	}
--- a/src/libfsm/edge.c
+++ b/src/libfsm/edge.c
@@ -43,7 +43,7 @@ fsm_addedge_any(struct fsm *fsm,
 
 	/* TODO: bulk add by symbol, presumably as a bitmap */
 	for (i = 0; i <= UCHAR_MAX; i++) {
-		if (!fsm_addedge_literal(fsm, from, to, (unsigned char) i)) {
+		if (!fsm_addedge_literal(fsm, from, to, (char)i)) {
 			return 0;
 		}
 	}
@@ -60,7 +60,7 @@ fsm_addedge_literal(struct fsm *fsm,
 	assert(from < fsm->statecount);
 	assert(to < fsm->statecount);
 
-	if (!edge_set_add(&fsm->states[from].edges, fsm->opt->alloc, c, to)) {
+	if (!edge_set_add(&fsm->states[from].edges, fsm->opt->alloc, (unsigned char)c, to)) {
 		return 0;
 	}
 
--- a/src/libfsm/minimise.c
+++ b/src/libfsm/minimise.c
@@ -265,7 +265,7 @@ build_minimised_mapping(const struct fsm *fsm,
 				if (next == NO_ID) {
 					continue;    /* 0 elements, skip */
 				}
-				should_gather_EC_labels = next & SMALL_EC_FLAG;
+				should_gather_EC_labels = (next & SMALL_EC_FLAG) != 0;
 				next = MASK_EC_HEAD(next);
 
 				if (env.jump[next] == NO_ID) {
--- a/src/libfsm/shortest.c
+++ b/src/libfsm/shortest.c
@@ -116,7 +116,7 @@ fsm_shortest(const struct fsm *fsm,
 
 			assert(v->state == e.state);
 
-			c = cost(u->state, v->state, e.symbol);
+			c = cost(u->state, v->state, (char)e.symbol);
 
 			/* relax */
 			if (v->cost > u->cost + c) {
--- a/src/adt/stateset.c
+++ b/src/adt/stateset.c
@@ -216,7 +216,6 @@ state_set_add(struct state_set **setp, const struct fsm_alloc *alloc,
 	size_t i;
 
 	assert(setp != NULL);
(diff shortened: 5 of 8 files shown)


src/adt/internedstateset.c
```diff
@@ -116,7 +116,7 @@ interned_state_set_pool_alloc(const struct fsm_alloc *a)
 	{
 		const unsigned long mask = CACHE_BUCKET_DEF_CEIL - 1;
 		struct cache_bucket *b = &res->cache.buckets[empty_i->hash & mask];
-		assert(b->iss = NO_ISS);
+		assert(b->iss == NO_ISS);
```
katef, line 119: oh!

src/libfsm/edge.c
```diff
@@ -43,7 +43,7 @@ fsm_addedge_any(struct fsm *fsm,
 
 	/* TODO: bulk add by symbol, presumably as a bitmap */
 	for (i = 0; i <= UCHAR_MAX; i++) {
-		if (!fsm_addedge_literal(fsm, from, to, (unsigned char) i)) {
+		if (!fsm_addedge_literal(fsm, from, to, (char)i)) {
```
katef, line 46: :+1:

src/libfsm/edge.c
```diff
@@ -60,7 +60,7 @@ fsm_addedge_literal(struct fsm *fsm,
 	assert(from < fsm->statecount);
 	assert(to < fsm->statecount);
 
-	if (!edge_set_add(&fsm->states[from].edges, fsm->opt->alloc, c, to)) {
+	if (!edge_set_add(&fsm->states[from].edges, fsm->opt->alloc, (unsigned char)c, to)) {
```
katef, line 63: :+1:

src/libfsm/minimise.c
```diff
@@ -265,7 +265,7 @@ build_minimised_mapping(const struct fsm *fsm,
 				if (next == NO_ID) {
 					continue;    /* 0 elements, skip */
 				}
-				should_gather_EC_labels = next & SMALL_EC_FLAG;
+				should_gather_EC_labels = (next & SMALL_EC_FLAG) != 0;
```
katef, line 268: I hate when compilers warn about this! but okay.

src/libfsm/shortest.c
```diff
@@ -116,7 +116,7 @@ fsm_shortest(const struct fsm *fsm,
 
 			assert(v->state == e.state);
 
-			c = cost(u->state, v->state, e.symbol);
+			c = cost(u->state, v->state, (char)e.symbol);
```
katef, line 119: :+1:

katef (approved):
Thanks!!

src/libfsm/minimise.c
silentbicycle: In this specific case it makes sense, because `SMALL_EC_FLAG` is the MSB, `should_gather_EC_labels` is a (signed) `int` being reluctantly used in place of a `bool` due to C89, whereas `next` is an `fsm_state_t`, an unsigned int, so it's implicit signed to unsigned conversion. I suppose we could have also used `!!`.

silentbicycle:
The UBSan build appears to be failing because of the warning about pointer arithmetic with a NULL pointer. clang-9 UBSan complains about it but the tests still pass, clang-10 UBSan complains and different behavior leads to test failure.

Locally editing the generated lexers so that the condition in the generated `_dynpush` function is is `f (t->a == NULL || (t->p == t->a + t->len))` fixes the UB and makes those tests pass, but when I change the codegen in `lx` and attempt to re-gen the lexers, I'm getting a build failure because the `lx->end.saved_col` field doesn't exist (perhaps in that specific build config).

hvdijk:
> I'm getting a build failure because the `lx->end.saved_col` field doesn't exist (perhaps in that specific build config).

Did you perhaps force files to be regenerated by removing the `lexer.c` files? If so, that doesn't tell the build system that `lexer.h` files also need to be regenerated. If you instead use `touch` to update the mtimes of the `lexer.lx` files, the build system should update both `lexer.c` and `lexer.h`, and the updated `lexer.h` should contain the declaration for `saved_col`.

silentbicycle:
The command I ran before building was:

    $ touch src/libre/dialect/*/lexer.lx

and that didn't seem to help. Retrying from a clean build didn't recover, and deleting the generated `lexer.[ch]` leads to a state where I can't build because they don't generate. It may be a bug in the makefile configuration? I also just confirmed that the `kmkf` submodule is up to date.

silentbicycle:
Manually running the command to regenerate the `lexer.h` header addresses it, so you're right about that part. There might be a missing dependency link between the lx file and the header.

It looks like the `lx.mk` makefile in `kmkf` is missing a dependency declaration for the generated .c file on the .h file, so the .h file can be stale. Fixing it locally got the build unstuck, so I'll pass that along.

silentbicycle:
No, it looks like the dependency is normally being created, but since I had to hardcode `CC=clang-10` to make some of the new UBSAN stuff work, rules in the build system that looked for exactly the `CC` named `"clang"` were no longer working, and that included some of the dependency metadata generation.

hvdijk:
There shouldn't be a dependency between the `.c` and `.h` files, either can be generated independent from the other. Either can be stale if only the other gets updated, and a dependency can only be one way, so you would still potentially have the same problem in the opposite direction if you try to address the problem that way. What `lx.mk` does instead is specify a dependency on the `.lx` file for both the `.c` and the `.h` file, so touching the `.lx` file like you did *should* cause both the `.c` and `.h` files to be regenerated. The `.o` file, however, should depend on both the `.c` and the `.h` files, and I see that that dependency is only present if you had a previous build, not if you tried with a clean build directory. Glad you got it working.

katef:
I'm going to merge this as-is, and we can deal with the lx issues separately.

PR #322, opened 2021-03-04
Less specific checks for compiler names

katef:
See https://github.com/katef/kmkf/issues/9


katef:
CI currently fails in sanitisers for unrelated reasons (presumably an update to clang on github's machines); I'm going to merge this first because it's convenient for testing with different versions of clang while we address those issues.

PR #324, opened 2021-03-17
Determinise with grouped edges

silentbicycle:
This reworks `fsm_determinise` to take advantage of label grouping in the new edgeset ADT. Casual benchmarking suggests this may be about 5x faster for my test data. There is a lot of room for further performance tuning, but this seems like a good waypoint.

Rather than analyzing every label independently, build a table of all of the `(label set)->to state` groups for each state set, then analyze the table to partition into non-overlapping groups, then add each set of edges all at once.

This adds a new ADT, "ipriq" (index priority queue), which is a heap-based priority queue whose cells are all indices into a separate structure. An opaque pointer to it and comparison callback are passed in. A new theft test exercises it.

### DIFF
--- a/include/adt/edgeset.h
+++ b/include/adt/edgeset.h
@@ -29,6 +29,39 @@ struct edge_ordered_iter {
 	uint64_t symbols_used[4];
 };
 
+enum edge_group_iter_type {
+	EDGE_GROUP_ITER_ALL,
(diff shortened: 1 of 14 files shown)


silentbicycle:
Made #325 to clean up some of the UBSan warnings, which are unrelated to this.

katef (approved):

PR #325, opened 2021-03-18
Address some UBSan warnings/errors related to numeric type conversion/overflow

silentbicycle:
This addresses several UBSan errors related to numeric types.

- Fix warnings about unsigned integer overflow in hash functions. That's defined, safe, and expected.
- Restructure a reverse iteration that counts down to not have potential UB around 0.
- Add a cast for a negated flag.

There are warnings remaining related to implicit conversions between `char` and `unsigned char`, but addressing them will require several interface changes. There is also warnings related to numeric type truncation in `parser.act` around `strtoul`, but those should be handled separately, especially since they will require the other parsers to be re-generated.

### DIFF
--- a/include/common/libfsm_common.h
+++ b/include/common/libfsm_common.h
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2021 Scott Vokes
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+#ifndef LIBFSM_COMMON_H
+#define LIBFSM_COMMON_H
+
+#if defined(__clang__)
+/* Newer versions of clang's UBSan emit warnings about *all* unsigned
+ * integer overflows. While they are defined behavior, overflow can
+ * cause bugs. This macro ignores them for a particular function.
+ * Overflow/rollover is expected in when hashing, for example. */
+#define SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()		\
+	__attribute__((no_sanitize("integer")))
+#else
+#define SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+#endif
+
+#endif
--- a/src/adt/internedstateset.c
+++ b/src/adt/internedstateset.c
@@ -13,7 +13,7 @@
 #include <adt/stateset.h>
 #include <adt/internedstateset.h>
 
-#include "common/libfsm_common.h"
+#include "common/check.h"
 
 #define ISS_BUCKET_DEF_CEIL 4
 #define CACHE_BUCKET_DEF_CEIL 32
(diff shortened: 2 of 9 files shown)


include/common/libfsm_common.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2021 Scott Vokes
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+#ifndef LIBFSM_COMMON_H
+#define LIBFSM_COMMON_H
```
katef, line 8: would you mind naming this something like check.h please? that way it doesn't invite more miscellaneous stuff in the same header

src/adt/internedstateset.c
```diff
@@ -168,6 +170,13 @@ interned_state_set_empty(struct interned_state_set_pool *pool)
 	return pool->empty;
 }
 
+SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+static unsigned long
+hash_state(fsm_state_t state)
+{
+	return PHI32 * (state + 1);
+}
+
```
katef, line 179: :+1:

src/adt/siphash.c
```diff
@@ -19,6 +19,9 @@
 #include <stdio.h>
 #include <string.h>
 
+/* Added to suppress UBSan warning for (expected) unsigned integer overflow. */
```
katef, line 22: (then there's no need for this comment, given the purpose of the header)

src/libre/ast_compile.c
```diff
@@ -865,7 +865,7 @@ comp_iter(struct comp_env *env,
 		re_flags = n->re_flags;
 
 		/* wouldn't want to reverse twice! */
-		re_flags &= ~RE_REVERSE;
+		re_flags &= ~(unsigned)RE_REVERSE;
```
katef, line 868: Aha!

katef (changes requested):
Thanks!

src/adt/siphash.c
silentbicycle: Okay. I was just being extra-explicit because it's modifying a reference implementation.

include/common/libfsm_common.h
silentbicycle: Updated in 2f154973.

silentbicycle:
The `Makefiles / bmake ubuntu DEBUG (push)` CI failure looks like a race condition in the build config (generating and installing the `.pc` files), which is being hit by `bmake -j 2`. That may need to be fixed via kmkf.

katef (approved):

katef:
I re-ran the CI jobs, looks like that's definitely an intermittent problem. How annoying!

PR #327, opened 2021-03-22
Overrides for a few more linguist things

katef:
There seems to be no way to test this; I think the github UI only shows classifications from `main`. So I guess I have to merge it and see if it does what I want.


katef:
update: It did what I want.

PR #336, opened 2021-04-12
Fix fsm_trim buffer overrun

silentbicycle:
theft found a potential buffer overrun when trimming the FSM produced by the regex "^(aaaaaeaaaabaa)?$". This regex led to an FSM where the start state, 13, was also an end, but the code in `fsm_trim` that determines how large the offset array should be only checked for the largest state with an edge to it -- since there was no edge to the start state, but it was reachable from the start, it was still live during trimming, and since it was an end state it was added to the queue for states with a path to an end state (using the reverse index). When popped from the queue, the state ID was greater than the offset array allocation, which led to an invalid read.

This is fixed by basing the offset array allocation size on the largest to-edge or end ID, _whichever is greater_.

Note that the attached test (`tests/trim/in4.fsm`) may not *fail* without this code change, but it has an invalid read 0 bytes past the allocation flagged in valgrind, and other sanitizers should also catch it. I found it because this input led to other assert failures later on.


katef (approved):
Thank you!

PR #341, opened 2021-04-17
Fixes to a few minor issues

sfstewman:
I'll start with a strong caveat: **I'm not at all sure about the AST fix**.

## Background

This started with the observation that the unanchored regepx `/|/` should match the string `xyz`.  However, it was compiling into the wrong DFA to match `xyz`.  Further, the empty (unanchored) regexp `//` was also compiling into the wrong DFA.

Looking back a bit, the DFA was incorrect because the NFA was incorrect.  The AST for both, after rewriting, was `EMPTY`, which seemed correct.

## Digging in, exonerating `ast_rewrite`

It looked like the issue was somewhere between when the AST was created and when it was translated into an NFA.  A usual suspect here is `ast_rewrite`, and indeed, when I disabled `ast_rewrite`. the unanchored expressions `//` and `/|/` gave the correct DFA.

However, `ast_rewrite` turned out to be an bystander, innocent of this crime if perhaps not entirely of pure character.  The initial AST for `/|/` is `AST(EMPTY|EMPTY)`.  Rewriting made two changes:

1. eliminate one of the `EMPTY` nodes; 
2. replace the resulting `AST(EMPTY)` tree with a single `EMPTY` node.  

Both of these changes seem entirely legitimate.  With `//`, `ast_rewrite` transforms `ALT(EMPTY)` into `EMPTY`, also legitimate.  So the issue seemed to be replacing the top-level ALT node with an EMPTY node.

Digging in a bit, `ALT(EMPTY)` and `EMPTY` have two potentially important differences:
1. In `ast_analysis`, the first/last flags were not set on an `EMPTY` node.
2. In `ast_compile`, the `EMPTY` node was not being linked to the global loop.

This is where I'm not at all sure what the correct solution is.  And, overcome by such ambiguity, I changed both.  This has fixed the DFA generation for `//` and `/|/`, but I can only hope that this solution is not worse than the original bug.

Please look it over carefully; the global loop bits always confuse me.  Feel free to disregard and fix it correctly!

## Unrelated minor fix

There's also a minor fix to the PCRE lexer that was causing the test suite to fail.

### DIFF
--- a/src/libre/ast_analysis.c
+++ b/src/libre/ast_analysis.c
@@ -451,6 +451,9 @@ assign_firsts(struct ast_expr *n)
 {
 	switch (n->type) {
 	case AST_EXPR_EMPTY:
+		set_flags(n, AST_FLAG_FIRST);
+		break;
+
 	case AST_EXPR_TOMBSTONE:
 		break;
 
@@ -523,6 +526,9 @@ assign_lasts(struct ast_expr *n)
 {
 	switch (n->type) {
 	case AST_EXPR_EMPTY:
+		set_flags(n, AST_FLAG_LAST);
+		break;
+
 	case AST_EXPR_TOMBSTONE:
 		break;
 
--- a/src/libre/ast_compile.c
+++ b/src/libre/ast_compile.c
@@ -372,6 +372,11 @@ decide_linking(struct comp_env *env,
 
 	switch (n->type) {
 	case AST_EXPR_EMPTY:
+		if ((n->flags & (AST_FLAG_FIRST|AST_FLAG_LAST)) == 0) {
+			return LINK_TOP_DOWN;
+		}
+		break;
+
 	case AST_EXPR_GROUP:
 		return LINK_TOP_DOWN;
 
--- a/src/libre/dialect/pcre/lexer.c
+++ b/src/libre/dialect/pcre/lexer.c
@@ -77,7 +77,7 @@ lx_pcre_dynpush(void *buf_opaque, char c)
 
 	assert(t != NULL);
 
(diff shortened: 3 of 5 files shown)


silentbicycle:
The global loop bits are _incredibly_ subtle and confusing.

In a working branch, I have a fuzzing harness to compare behavior against PCRE. It's a work in progress, and currently focusing on the group match behavior. Along with reviewing the code changes, I will cherry-pick this into the branch later this week to see whether it finds or fixes inconsistencies.

silentbicycle:
The code changes look reasonable to me. I will think through the implications and generate test inputs as time allows.

silentbicycle (changes requested):
Sorry it took so long for me to get to this. Testing didn't turn up any new issues so far.

When you address the lexer conflict, could you also please add a regression test for `/|/` matching `"xyz"`? Otherwise looks good to me.

sfstewman:
Rebased onto katef:kate/ci-retest-pcre-suite

I dropped 9aa46465e093031dbc385265dc4007b1b608b360, which @katef had independently incorporated as 8d67371880540af5f1c9e8e0ca4e153f22b426d6

sfstewman:
Adds two regression tests for the PCRE dialect:
- unanchored `/|/` matches 'xyz'
- unanchored `//` matches 'xyz'

katef:
Oh sorry, I meant can you base this against `main` please? I'd like to merge it before the kate/ci-retest-pcre-suite branch

sfstewman:
Okay.  Rebased the relevant bits (including the fix for the lx ambiguity) onto main.

katef (approved):
Thank you!!

PR #342, opened 2021-04-19
README.md - mention bmake and how to run the test suite

federicomenaquintero:


README.md
```diff
@@ -51,8 +54,17 @@ Building depends on:
...
+ * https://github.com/silentbicycle/theft
    Tests are currently based on the libtheft 0.4.2 API.
 
 Ideas, comments or bugs: kate@elide.org
 
+### Running the test suite
+
+You need `sid` installed; it is a parser generator.  You can build it
+with `pmake` / `bmake` from https://github.com/tendra/tendra/tree/main/sid 
+and then just copy the binary to somewhere in your PATH.
```
katef, line 66: This isn't actually true! It's just an unfortunate side effect of the timestamps from checkout. The generated code from SID is in the repo, so perhaps we could say something like touch the generated parser.c/.h files _or_ install SID.

README.md
```diff
@@ -51,8 +54,17 @@ Building depends on:
 
 Fuzzing depends on the theft property-based testing library:
 
- * https://github.com/silentbicycle/theft  
+ * https://github.com/silentbicycle/theft
```
katef, line 57: The double space here (it means `<br/>`) was intentional!

README.md
federicomenaquintero: Hehe, ok - I couldn't make it work by just touching those files.

README.md
federicomenaquintero: Today I learned!  I thought I had left some garbage there in my editor :-P

README.md
federicomenaquintero: Uh, upon looking at `.github/workflows/make.yml` more closely - would you recommend running the tests with `SID=true` as there?

katef (approved):
Thank you!

README.md
katef: oh sorry, I missed this comment. That's a hack I'm not sure I want to encourage. it's so confusing. I'm not sure what the best thing is to suggest, here.

PR #343, opened 2021-04-20
Import Cargo, first steps towards porting to Rust 🦀

katef:
Here I'm introducing Cargo to the build. It's driven from the existing makefiles. This marks the start of porting to Rust by a piecemeal approach, so we'll have a hybrid codebase until that's completed.

I've added an `fsm_noop` function written in Rust (called from `fsm_new`), just to prove the linking here works.

Currently the makefiles build libfsm.a from a single partially-linked libfsm.o. That single .o file is made by `ld -r -o libfsm.o *.o`, effectively bundling together all the various .o files.

In this PR I'm adding the Cargo-generated .a file to that list of sources for `ld -r`. I think this depends on ld being able to see inside a .a file when invoked with -r for partial linking. This seems to work okay, at least on the ELF and Mach-O systems available in CI. Those are all I'm interested in for the moment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
-build/
+/build
+/target
+Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Kate F <kate@elide.org>"]
 edition = "2018"
 
 [lib]
+crate-type = ["staticlib"]
 path = "src/libfsm/lib.rs"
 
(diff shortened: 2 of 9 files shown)


data-man:
Really porting to Rust?
Bad news.

katef:
> Really porting to Rust?
> Bad news.

I'm unsure. at least for the moment, it's more work than I can take on, regardless of whether it's a good idea or not. So this PR isn't going anywhere for a while yet.

PR #344, opened 2021-04-20
Move libfsm/start.c to Rust

federicomenaquintero:
This depends on #343.

First we replace the bitfields in a couple of structs with `stdbool.h`, so we can have the same ABI as Rust's `bool` type.

Then we make parallel structs in Rust for `struct fsm_state` and `struct fsm`.  We'll keep these in sync with the C structs while the internals get rustified.

Then we port `fsm_clearstart`, `fsm_getstart`, `fsm_setstart` to Rust, and remove `start.c`.

I probably totally botched the Makefiles to add `-lpthread -ldl`, needed by Rust libraries, but I need some help from someone who knows pmake :)


federicomenaquintero:
The last commits I've pushed (`fsm_addstate()` onwards) are a little refactoring of some of the C code.  My goal here is to isolate the use of `fsm->statealloc` to `fsm.c`, which should hopefully make it easier to turn the states array into a Rust `Vec`.

Those commits are useful outside of a Rust port.  Please tell me if you'd rather have me submit a separate pull request for them, or if you'd prefer to cherry-pick them or whatever.

PR #348, opened 2021-04-26
fsm_print_fsm: add option for printing grouped edges, add CLI flag.

silentbicycle:
Add a flag to `struct fsm_options`, `group_edges`, which tells fsm to print edges in a group whenever the output method supports it.

This can make fsm output much easier to read with FSMs from unanchored regexes, which often have hundreds of "...and everything else 0x00 - 0xFF goes back to state X" edges.

Without the `-g` flag, this output would be 1798 lines long:

    0; 1; 2; 3; 4; 5; 6; 7;

    0  ->  0 "\x00" .. "a", "c" .. "\xff";
    0  ->  1 "b";
    1  ->  0 "\x00" .. "`", "c" .. "\xff";
    1  ->  1 "b";
    1  ->  2 "a";
    2  ->  0 "\x00" .. "a", "c" .. "m", "o" .. "\xff";
    2  ->  1 "b";
    2  ->  3 "n";
    3  ->  0 "\x00" .. "`", "c" .. "\xff";
    3  ->  1 "b";
    3  ->  4 "a";
    4  ->  0 "\x00" .. "a", "c" .. "m", "o" .. "\xff";
    4  ->  1 "b";
    4  ->  5 "n";
    5  ->  0 "\x00" .. "`", "c" .. "\xff";
    5  ->  1 "b";
    5  ->  6 "a";
    6  ->  0 "\x00" .. "a", "c" .. "r", "t" .. "\xff";
    6  ->  1 "b";
    6  ->  7 "s";
    7  ->  7 ?;

    start: 0;
    end:   7;

Note that this change does not update the man page, any other output modes (it would be nice to support this in `dot` later), or make fsm able to read that input back in yet.

### DIFF
--- a/include/fsm/options.h
+++ b/include/fsm/options.h
@@ -47,6 +47,11 @@ struct fsm_options {
 	 */
 	unsigned int always_hex:1;
 
+	/* boolean: true indicates to group edges with a common destination in output,
+	 * when possible, rather than printing them all individually.
+	 */
+	unsigned int group_edges:1;
+
 	/* for generated code, what kind of I/O API to generate */
 	enum fsm_io io;
 
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -106,7 +106,7 @@ static void
 usage(void)
 {
 	printf("usage: fsm [-x] {<text> ...}\n");
-	printf("       fsm {-p} [-l <language>] [-aCcwX] [-k <io>] [-e <prefix>]\n");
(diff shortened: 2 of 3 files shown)


sfstewman:
My $0.02: This is *awesome* and I love it.

silentbicycle:
This flag _looks_ redundant with `consolidate_edges` (and, really, it is), but that one is set by default in `lx` and `re`. Until we can also read in that output with `fsm`, their output printing grouped edges will break most of the tests.

Once `fsm` can read that format, the `-g` flag should probably go away.

katef (approved):
Thank you!

PR #349, opened 2021-04-27
Rename `fsm_glushkovise` to `fsm_rename_epsilons` throughout.

silentbicycle:
"glushkovise" is a bit misleading here, since it isn't the full Glushkov algorithm for construction, just removing epsilon edges as a first pass before determinisation.


silentbicycle:
Sanitizer failure: I can't reproduce this locally (but I'm not using macos), but the issue appears to be related to `fsm_glushkovise` (now `fsm_remove_epsilons`)'s updating of the capture action metadata being incorrect and `fsm_determinise` tripping up on it later on. That's actually what I'm currently trying to fix on a different branch, so I am going to continue with that for now and then update this later.

katef (approved):

PR #356, opened 2021-05-24
trim: Fix potential buffer overrun.

### LINKED ISSUE #355: "error: Success" from fsm_trim
```
; cat /tmp/ha-target.fsm 
0; 1; 2; 3; 4; 5;

0  ->  1;
0  ->  1 "e";
1  ->  2 "b";
2  ->  3 "g";
2  ->  4 "h";
3  ->  0 "h";
4  ->  0;
4  ->  1 "h";
5  ->  1 "a";
5  ->  3 "h";

start: 5;
end:   3;
```
gives:
```
; fsm -t trim -capl dot < /tmp/ha-target.fsm
fsm_op: Success
```

silentbicycle:
This is related to the bug that was fixed previously in
b71637d9fef43f35376ab5c14d90e6c33fbf5b23. (See description.)
This could be triggered by the following input:

    0; 1; 2; 3; 4;
    0  ->  1 "g";
    1  ->  2 "b";
    2  ->  1;
    3  ->  3 "h";
    4  ->  0 "b";
    start: 4;
    end:   0;

The other fix addresses the case where the end state was larger, this
version is with the start state. Strictly speaking, the offset length
can be the max of `[max_end, max_to, start]` + 1, but it's simpler to
just use the state count, rather than attempting to save memory by
allocating a few less states if the last few are unreachable.

Fixes #355.


katef (approved):
That makes sense, and I agree that it's not a situation that makes sense to optimise for. I don't expect to produce FSMs where it's significant that the tail of the state array is unreachable.

Thank you!

PR #363, opened 2021-06-24
add portable overflow checking multiply

moon-chilled:
This gets it building under tcc


katef (approved):
I totally missed this! Thank you!

PR #368, opened 2021-10-22
DFA group captures -- Explicit AST stack and `fsm_generate_matches`

silentbicycle:
This adds the next layer of changes preparing for handling group capture in DFAs:

- Add `fsm_generate_matches`, which will be used for comparison testing later.
- Replace the C-stack `ast_compile.c` implementation with one using an explicit data stack. This will be necessary to properly handle captures nested in ALT nodes.
- Split the start and end nodes into an `_inner` and `_outer` pair, to explicitly track what occurs inside of match group 0.
- Several other minor clarifications, bug fixes, and performance improvements.

See the commit logs for details.


src/libfsm/endids.c
```diff
@@ -129,7 +129,7 @@ fsm_endid_free(struct fsm *fsm)
 static unsigned long
 hash_state(fsm_state_t state)
 {
-	return fsm_hash_id(state);
+	return (unsigned long)fsm_hash_id(state);
```
katef, line 132: `fsm_hash_id()` returns `uint64_t`. the minimum maximum for `unsigned long` is 32-bit. this'll bite us when compiling libfsm to wasm. since you're introducing a change around these variables anyway, let's make all these hash variables uint64_t now

src/libfsm/fsm.c
```diff
@@ -127,6 +127,7 @@ fsm_move(struct fsm *dst, struct fsm *src)
 
 	dst->states     = src->states;
 	dst->start      = src->start;
+	dst->hasstart   = src->hasstart;
```
katef, line 130: oh wow. surprised this was missing! thanks.

src/libfsm/gen_internal.h
```diff
@@ -0,0 +1,132 @@
...
+sfs_step_self_ref(struct gen_ctx *ctx, struct gen_stack_frame *sf);
+
+static bool
+sfs_leaving_state(struct gen_ctx *ctx, struct gen_stack_frame *sf);
+
+static bool
+grow_buf(struct gen_ctx *ctx);
+
+static bool
+grow_stack(struct gen_ctx *ctx);
```
katef, line 130: why the static prototypes there?

src/libfsm/minimise.c
```diff
@@ -386,64 +387,6 @@ dump_ecs(FILE *f, const struct min_env *env)
...
 
 #define PARTITION_BY_END_STATE_DISTANCE 1
-#if PARTITION_BY_END_STATE_DISTANCE
-/* Use counting sort to construct a permutation vector -- this is an
- * array of offsets into in[N] such that in[pv[0..N]] would give the
- * values of in[] in ascending order (but don't actually rearrange in,
- * just get the offsets). This is O(n). */
-static unsigned *
-build_permutation_vector(const struct fsm_alloc *alloc,
-    size_t length, size_t max_value, unsigned *in)
```
katef, line 396: good call to move this out

src/libfsm/trim.c
```diff
@@ -224,7 +231,34 @@ mark_states(struct fsm *fsm, enum fsm_trim_mode mode,
 	}
 
 	/* Sort edges by state they lead to, inverting the index. */
-	qsort(edges, edge_count, sizeof(edges[0]), cmp_edges_by_to);
```
katef, line 227: very nice

src/libre/ast_compile_internal.h
```diff
@@ -0,0 +1,162 @@
+#ifndef AST_COMPILE_INTERNAL_H
+#define AST_COMPILE_INTERNAL_H
```
katef, line 2: if this stuff is really only used by ast_compile.c, i'd rather we write it directly in that .c file. likewise for the other *_internal.h headers. having a separate header means i'd need to read this at the same time as the .c file in order to understand the structs and such when they're only ever used there, but also hints that other .c files might include these too, but we're not introducing an internal api here

src/libre/ast_compile_internal.h
```diff
@@ -0,0 +1,162 @@
...
+static int
+eval_GROUP(struct comp_env *env, struct comp_stack *stack);
+static int
+eval_ANCHOR(struct comp_env *env, struct comp_stack *stack);
+static int
+eval_SUBTRACT(struct comp_env *env, struct comp_stack *stack);
+static int
+eval_RANGE(struct comp_env *env, struct comp_stack *stack);
+static int
+eval_TOMBSTONE(struct comp_env *env, struct comp_stack *stack);
```
katef, line 160: i wouldn't include static prototypes unless you need them (e.g. for mutually recursive functions)

katef (changes requested):

src/libfsm/endids.c
silentbicycle: Sure, will do, and intentionally using `uint64_t` rather than `unsigned long` will remove some other annoyances elsewhere.

src/libfsm/gen_internal.h
silentbicycle: Habit. If we don't want to have them when they're not strictly necessary for forward references, I can remove them.

src/libre/ast_compile_internal.h
silentbicycle: Okay. I specifically moved them into their own files so I could pull up the struct definitions, prototypes, etc. alongside the code in another buffer, not because I expect anything else to call them.

include/adt/pv.h
```diff
@@ -0,0 +1,106 @@
...
+		goto cleanup;
+	}
+	counts = f_calloc(alloc, max_value + 1, sizeof(*out));
+	if (counts == NULL) {
+		goto cleanup;
+	}
+
+	/* Count each distinct value */
+	for (i = 0; i < length; i++) {
+		const unsigned *vp = (const unsigned *)&raw8[i*struct_size + offset];
```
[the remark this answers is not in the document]
silentbicycle: It checks that the struct size is >= `sizeof(unsigned)` and that the offset is a multiple of `sizeof(unsigned)`, but does not check the alignment of the multiplied struct size plus the offset. Do you have any suggestions for how to structure that without it being either a lot less clear or adding lots of extra runtime overhead (e.g. a callback that casts to a structure and returns the field as an unsigned)?

src/libre/ast_compile_internal.h
silentbicycle: These are forward references.

src/libfsm/endids.c
silentbicycle: Updated in a128cb27756bcf505b1a87eef7cb2a5ea82b8766.

src/libre/ast_compile_internal.h
silentbicycle: Updated in 4b1f511e523054ce4f892dbcf07728217a49ec50.

katef:
CI fails when fetching the pcre2 test suite because pcre's ftp server no longer exists. I've switched this to fetch from a different source in #369; please bring that commit here when #369 is merged to main so we can see CI pass.

katef (approved):
Thank you

include/adt/pv.h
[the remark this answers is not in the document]
katef: no, i don't have any ideas about that

silentbicycle:
`main` has been merged into the integration branch, and the PR branch has been rebased to include its changes.

katef (approved):
Thank you Scott!

PR #373, opened 2022-01-21
DFA group captures -- Minimise with label sets

silentbicycle:
This is probably the last PR against the integration branch before a large chunk of new capture processing and test code that goes together.

- Add a `-C` flag to `re` that disables the comment in the FSM output, since generating those can be expensive enough to slow down certain kinds of testing. It was much more significant without label sets, but it still adds up. I am not attached to using `-C` specifically for this, so let me know if you'd rather it go under a different letter or some kid of flag option interface. `-C` may be useful later for explicitly enabling/disabling captures, since at least initially we will not be able to support certain constructs in PCRE when captures are in use.
- Remove the `#ifdef`'d out non-grouped-label implementation for edge sets, since the codebase no longer builds with that implementation. Several important operations now depend on the grouped edge labels for performance.
- Use information about label sets to speed up minimisation: When checking the destination EC for a particular label's edge, also note which other labels lead to the same state, because they have already been checked and can be skipped later. This eliminates highly redundant processing in the innermost loop of minimisation when several labels lead to the same state, which is very common in complex DFAs.


katef (approved):

src/re/main.c
```diff
@@ -663,9 +663,10 @@ main(int argc, char *argv[])
 	{
 		int c;
 
-		while (c = getopt(argc, argv, "h" "acwXe:k:" "bi" "sq:r:l:F:" "upMmnfxyz"), c != -1) {
+		while (c = getopt(argc, argv, "h" "aCcwXe:k:" "bi" "sq:r:l:F:" "upMmnfxyz"), c != -1) {
```
katef, line 666: i don't have much preference about `-C`, as long as it matches fsm(1) (which currently uses `-C` as i'm sure you saw)

PR #374, opened 2022-03-31
Add basic fuzzer (using libfuzzer), bugfix

silentbicycle:
Add a fuzzer (`build/fuzz/fuzzer`), which generates randomized input strings, attempts to build them as a PCRE regex, and if accepted it determinises, minimises, and generates C code for the resulting FSM.libfuzzer's coverage/branch instrumentation guides input generation.

I will build on the fuzzer for comparing behavior with PCRE (especially capture group behavior), but this already found one bug, and regression tests and a fix are included.

This could probably be better integrated into the build system. It shouldn't be possible to try to build without setting `FUZZER` and using `CC=clang*`. There was already a `fuzz` makefile target (which builds and runs the theft tests), so I named it `fuzzer`. Should we rename the theft `fuzz` target to `prop`, since it's property tests?


src/libre/ast_analysis.c
```diff
@@ -368,7 +368,6 @@ analysis_iter_anchoring(struct anchoring_env *env, struct ast_expr *n)
 				struct ast_expr *doomed = n->u.alt.n[i];
 				n->u.alt.n[i] = ast_expr_tombstone;
 				ast_expr_free(env->pool, doomed);
-				continue;
```
katef, line 371: :D

katef (approved):

Issue #366, opened 2021-09-18
Clarification needed README instructions re: pmake/bmake on mac


xandhiller:
Regarding the instructions in the README

> To build and install:
> ```
> ; pmake -r install
> ```
> To build on Linux, you may need to install `bmake`, and use it instead of `pmake` in the commands mentioned here.

I had to use `bmake` on Mac too, might want to append this instuction here with that info. Took me ten minutes to try that and I'd like to save someone else the same trouble.

katef:
Thanks! You're right, I definitely should explain this more clearly. The trouble is there are a bunch of different BSD make packages around, and for some bewildering reason they all go by different names. And the makefiles _should_ work with all of them. But the name is whatever the package maintainer gave it! I'm really not sure how to explain that in the readme. I guess I just need the same section talking about bmake per-OS for every project.

katef:
What do you think?
https://github.com/katef/libfsm/pull/376

PR #376, opened 2022-05-02
Clarifaction on bmake

katef:
Suggested by #366. If this is good, I'll copy over the same text for all my stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,6 @@ Getting started:
  * [Compilation phases](doc/tutorial/phases.md) for typical applications
    which compile regular expressions to code.
 
-Lexing is the process of categorising a stream of items by their spellings.
-The output from this process is a stream of tokens, each of a specific lexeme
-category, which are most commonly input to a parser responsible for asserting
-the order of these tokens is valid.
-
-lx is an attempt to produce a simple, expressive, and unobtrusive lexer
-generator which is good at lexing, does just lexing, is language independent,
-and has no other features.
-
 You get:
 
  * libfsm — library for manipulating FSM (NFA and DFA)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,28 +19,52 @@ You get:
...
-To build and install:
+### Building
 
-    ; pmake -r install
+To build and install:
 
-To build on Linux, you may need to install `bmake`, and use it instead
-of `pmake` in the commands mentioned here.
+    ; bmake -r
+    ; bmake -r install
```
sfstewman, line 35: I'm tempted to say that you're introducing the term `bmake` without explaining it here, but I'm not sure that there's a better way without making it really wordy.

Maybe add a note about `bmake`, something like:
> To build `libfsm` requires a BSD-style make, which we'll refer to as `bmake` here to clearly distinguish it from GNU make.  See the note about `bmake` below for how to choose/obtain a BSD-style make.

README.md
```diff
@@ -28,28 +19,52 @@ You get:
...
+ * If you use Linux and you have a package named pmake, use that.
+   It's the same thing.
+   Some package managers have bmake packaged under the name pmake.
+   I don't know why they name it pmake.
+ * Otherwise if you use MacOS and you only have a package named bsdmake, use that.
+   It's Apple's own fork of bmake.
+   It should also work but it's harder for me to test.
+ * If none of these are options for you, you can build bmake from source.
+   You don't need mk.tar.gz, just bmake.tar.gz.
+   https://www.crufty.net/help/sjg/bmake.html
```
sfstewman, line 60: I like the detail here; there's a lot.  I was initially worried about it being confusing, but re-reading it, I don't think it is.

One thing you may clarify is that sjg's `bmake` always works (as long as it compiles).  I've run into problems with `bmake/pmake/bsdmake` on Linux and macOS before.

man/lx.1/lx.1.xml
```diff
@@ -86,6 +86,12 @@
 			and outputs a program which implements that analyser.
 			The specification language is documented by &lx.5;</para>
 
+		<para>Lexing is the process of categorising a stream of items
+			by their spellings. The output from this process is
+			a stream of tokens, each of a specific lexeme category,
+			which are most commonly input to a parser responsible
+			for asserting the order of these tokens is valid.</para>
+
```
sfstewman, line 94: The one thing I feel is missing is that lexer output is often not just a token but a token and the specific identifier/spelling associated with it.

Also, this is maybe not the place for it, but I think an example would help to clarify this enormously.  There are a lot of terms here (eg: tokens, lexeme category) , and I'm not sure if they're explained elsewhere.  Something like:

> As an example, a lexer for a C-like language would take the input
> `direction = "north";` and output the tokens 
> `IDENTIFIER EQUALS STRING_LITERAL SEMICOLON`, 
> where `IDENTIFIER` and `STRING_LITERAL` would have the associated spellings `direction` and `north`, respectively.

sfstewman (approved):
Overall this is a nice improvement!  I had a couple of comments that I think might clarify things.  Feel free to disregard.

README.md
katef: I think this is overkill, but moreso it gets in the way

README.md
katef: Added in 2bee00a, and I numbered the list items.

PR #384, opened 2022-09-01
Handle unsatisfiable subexpressions like `^a($b)c$` at the right step.

silentbicycle:
We were previously replacing the subtree with an `AST_EXPR_EMPTY` node
when a repeat node had an unsatisfiable subexpression, but that was the
wrong time to do it, because other steps assume that all
`AST_EXPR_EMPTY` nodes have been filtered out by the end of AST
rewriting.

Instead of changing the node type, just set the `AST_FLAG_UNSATISFIABLE`
flag on the AST node and continue. (If the min count is non-zero, it
still reports the whole AST unsatisfiable and halts, but a zero count
can be ignored.) Later on, check for that flag and skip the subtree.

I think this was previously asymptomatic, but changes to address a
different unsatisfiability bug with ALT subexpressions made it possible
for `tests/native/in37.re` to trigger an assert failure.

### DIFF
--- a/src/libre/ast_analysis.c
+++ b/src/libre/ast_analysis.c
@@ -311,6 +311,13 @@ analysis_iter_anchoring(struct anchoring_env *env, struct ast_expr *n)
 			 */
 			assert(child->type != AST_EXPR_EMPTY);
 
+			/* If a subtree is unsatisfiable and can be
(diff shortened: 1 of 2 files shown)


katef (approved):

katef:
Thanks!

PR #388, opened 2022-11-22
libre: Rework AST analysis and AST -> NFA compilation. Add several new PCRE-anchor regression tests.

silentbicycle:
This substantially reworks libre's AST analysis and AST -> NFA
compilation phases in order to correctly handle a bunch of cases that
were found while fuzzing and comparing behavior with PCRE. Most
have to do with interactions between start/end anchors and other features.

These changes came from a development branch adding support for capture
groups, but the changes here are independent and could be reviewed
separately. There are also many integration tests related to incorrect
capture bounds, but those will be merged with the capture changes.

- In general, frame analysis so that its results can be understood by
  looking at re's tree printout (`-ltree`), rather than by studying logs
  from lots of intermediate steps. Once flags are set they are never
  removed. This makes interactions leading to unsatisfiability easier
  to understand, because the whole subtree is marked with X and it
  either bubbles up or is ignored by particular parent nodes.

  Special case: A repeat node which has had its max reduced to 0, which
  usually happens due to an unsatisfiable subtree, has its
  AST_FLAG_CAN_CONSUME flag ignored rather than cleared.

- Add several AST flags, and add support for them in the tree
  printout.

- During the first analysis pass, do a bunch of extra checks and
  flag subtrees with AST_FLAG_UNSATISFIABLE.

- Add forward and backward passes over the AST that use details from
  analysis up to that point to detect many cases where interactions
  between anchoring and other features make subtrees unsatisfiable. For
  example, any subtree that always consumes input but is preceded by an
  end anchor gets AST_FLAG_UNSATISFIABLE. Unsatisfiable subtrees in an
  AST_EXPR_ALT node can be skipped, and in certain other cases they can
  also be ignored rather than causing the entire AST to be rejected.

- Add a flag `ast_expr.u.anchor.is_end_nl` which notes which end
  anchors do/do not accept a trailing newline. This is necessary
  to distinguish between `$` and `\z` in PCRE.

- The node linking types (`LINK_TOP_DOWN, LINK_GLOBAL,
  LINK_GLOBAL_SELF_LOOP`) are now mutually-exclusive, which
  simplifies the link time analysis.

- Add `intern_end_nl`, which is like `intern_end_any_loop` but adds a
  single state with both epsilon and literal '\n' matching edges to the
  global end, if not already present. This creates a single global
  destination for PCRE's '\n'-matching `$` operator. Previously, `$` was
  rewritten to a `\n?$` subtree inside the PCRE parser, but since PCRE
  accepts redundant `$` anchors and `$` anchors precending `^`, ensuring
  that the resulting NFA only matches a trailing newline *once* becomes
  very difficult. Fuzzing found several pathological cases like `$^$`
  that led to an incorrect result, but `intern_end_nl` completely
  eliminates the whole issue of tracking how many implicit newlines have
  been accepted, and the parser just creates an end-nl anchor instead
  of an end anchor as necessary.

- PCRE should create a global match group 0 surrounding the parse. In a
  subsequent PR, the interned start/end states will be split
  into (start/end)_inner and _outer, representing what part of the
  matched input is within the bounds of match group 0 -- input matched
  by the unanchored start and end loops are not included. This also
  changes the parser's action file to start the group counter from 0
  rather than 1.

- Fix a bug where the last node in an AST_EXPR_CONCAT that was
  unsatisfiable but also nullable incorrectly caused the entire CONCAT
  node to become disconnected, rather than skipping the last node and
  linking to the top-down end state.

- Add a line to the other dialects' parsers noting that they do not
  use the new `ast-make-anchor-end-nl` action.

- Add a flag to `re_dialect` for PCRE noting that it has special
  handling for `$` matching a trailing newline. I don't know if you
  want to include this behavior in `RE_NATIVE`, because it leads to
  numerous ugly special cases.

- Stop handling direct linkage to the global `env->start`/`env->end`
  inside the handler for `AST_EXPR_ANCHOR` nodes, because this leads to
  an incorrect result when anchors are redundant and/or out of order. It
  is technically legal in PCRE to use `$^`, `$$$^^^`, or `$^$^$` rather
  than `^$`.

- Add `ast_node_type_name`, to print node types when logging.

- Rename a couple 'have_*' flags to 'has_*'. Those may have been a typo.

- Fix incorrect pcre-flags tests. Since these are run with `-b`, for
  `RE_ANCHORED`, they should behave like PCRE when both `PCRE2_ANCHORED`
  *and* `PCRE2_ENDANCHORED` are set. (I suspect the latter was missed.)
  In that case, the trailing `$` anchor does not match a trailing '\n'.

- Add tests/pcre-anchor/ tests 58-79. These check various
  anchoring-related edge cases. Most were found while fuzzing to compare
  capture behavior between PCRE and libfsm. A few are asymptomatic
  without checking the resulting capture bounds, but most have to do
  with things incorrectly being marked / not marked as unsatisfiable due
  to interactions between anchors and other features.


katef:
> Previously, `$` was rewritten to a `\n?$` subtree

Noting that this PR replaces the subtree work introduced in #334, and that some of the tests changed on that PR are now changed back to their original behaviour.

src/libre/ast_compile.c
```diff
@@ -727,7 +681,7 @@ comp_iter(struct comp_env *env,
 
 	case AST_EXPR_CONCAT:
 	{
-		fsm_state_t base, z;
+		fsm_state_t base;
```
katef, line 684: I think this can go

src/libre/ast_analysis.c
```diff
@@ -153,77 +252,326 @@ analysis_iter(struct ast_expr *n)
...
+	case AST_EXPR_ALT: {
+		for (size_t i = 0; i < n->u.alt.count; i++) {
+			struct ast_expr *child = n->u.alt.n[i];
+			set_flags_subtree(child, flags);
+		}
+		break;
+	}
+
+	case AST_EXPR_REPEAT:
+		set_flags_subtree(n->u.repeat.e, flags);
```
katef, line 527: Is the fall-through here intentional? It doesn't look like it! If it's not, I'm surprised fuzzing didn't catch this.

src/libre/print/tree.c
```diff
@@ -57,6 +57,11 @@ fprintf_flags(FILE *f, enum ast_flags flags)
 	PR_FLAG(LAST,          "L");
 	PR_FLAG(UNSATISFIABLE, "X");
 	PR_FLAG(NULLABLE,      "0");
+	PR_FLAG(ANCHORED_START,"^");
+	PR_FLAG(ANCHORED_END,  "$");
+	PR_FLAG(END_NL,        "N");
+	PR_FLAG(CAN_CONSUME,   "c");
+	PR_FLAG(ALWAYS_CONSUMES, "C");
```
katef, line 64: Noting that we don't currently print `(struct ast_expr)->flags` from the corresponding AST rendering in src/libre/print/dot.c
(but we should!)

src/libre/ast.c
```diff
@@ -893,3 +893,22 @@ ast_make_expr_named(struct ast_expr_pool **poolp, enum re_flags re_flags, const
 	return NULL;
 }
 
+const char *
+ast_node_type_name(enum ast_expr_type t)
+{
+	switch (t) {
+	default:
+		return "<match fail>";
```
katef, line 901: convention is for `return "?";` in this repo

src/libre/ast_analysis.c
```diff
@@ -64,38 +86,95 @@ analysis_iter(struct ast_expr *n)
...
+			LOG(3 - LOG_FIRST_ANALYSIS,
+			    "%p: any_can_consume -> setting CAN_CONSUME\n", (void *)n);
+			set_flags(n, AST_FLAG_CAN_CONSUME);
+		}
+
+		/* FIXME: This should not be invalidated by unsatisfiable nodes without the
+		 * AST_FLAG_ALWAYS_CONSUMES flag set, but since unsatisfiability derives
+		 * from needing to consume input in a way that contradicts anchoring, it
+		 * may not be possible to produce a case where an alt subtree is
+		 * unsatisfiable but does not set AST_FLAG_ALWAYS_CONSUMES. */
```
katef, line 171: Would it make sense to reject compilation for this situation, just in case it does occur? I'm taking the "it may not be possible" part to mean that we don't have test coverage for this situation.

src/libre/ast_analysis.c
```diff
@@ -139,8 +228,18 @@ analysis_iter(struct ast_expr *n)
...
+		if (res != AST_ANALYSIS_OK) { return res; }
+		res = analysis_iter(n->u.subtract.b);
+		if (res != AST_ANALYSIS_OK) { return res; }
+		enum ast_flags a_consume_flags = n->u.subtract.a->flags &
+		    (AST_FLAG_CAN_CONSUME | AST_FLAG_ALWAYS_CONSUMES);
+		if (a_consume_flags != 0) {
+			set_flags(n, a_consume_flags);
+		}
 		break;
+	}
```
katef, line 242: great!

src/libre/ast_compile.c
```diff
@@ -28,14 +28,17 @@
...
+#endif
+
 enum link_side {
 	LINK_START,
 	LINK_END
 };
 
 /*
  * How should this state be linked for the relevant state?
- * Note: These are not mutually exclusive!
```
katef, line 38: nice

src/libre/parser.act
```diff
@@ -1080,7 +1090,7 @@
 		act_state->overlap = overlap;
 		act_state->poolp   = &ast->pool;
 
-		act_state->group_id = 1;
+		act_state->group_id = 0;
```
katef, line 1093: There's a comment around line 127 near the top of this file about the numbering for `group_id`; if we're numbering from 0 now, I think that needs updating. (Or I don't follow what the 0 here means)

src/libre/re.c
```diff
@@ -40,7 +40,7 @@ re_dialect(enum re_dialect dialect)
 		{ RE_LITERAL, parse_re_literal, 0, RE_SINGLE | RE_ANCHORED },
 		{ RE_GLOB,    parse_re_glob,    0, RE_SINGLE | RE_ANCHORED },
 		{ RE_NATIVE,  parse_re_native,  0, 0                       },
-		{ RE_PCRE,    parse_re_pcre,    0, 0                       },
+		{ RE_PCRE,    parse_re_pcre,    0, RE_END_NL               },
```
katef, line 43: Should we have `RE_END_NL` for the native dialect too? I'm not sure.

src/libre/ast_analysis.c
silentbicycle: Oh, yikes, good catch. Those happen to be safe though coincidence, because all three fields in the union occur at the same offset and repeatedly setting the same flags has no effect, but that's definitely a bug. Fixed in 4cfe6e34.

src/libre/ast_compile.c
silentbicycle: Indeed. Removed in 18dbc414.

src/libre/print/tree.c
silentbicycle: We probably should, but I don't tend to use the dot view much -- since it doesn't always collapse down multiple edges between nodes, it's easy to produce dot views where nodes with 254 edges between them make the graph very large and difficult to read. If we rework dot output to handle that sort of thing better, we should include the flags at the same time.

src/libre/ast.c
silentbicycle: Fixed in 676e61d0.

src/libre/ast_analysis.c
silentbicycle: I don't think it's possible here because it relies on information that won't be available until an analysis pass that has to run after this one completes, but I also haven't been able to find an input that produces incorrect results due to that situation.

I think rejecting it would be for the best, but a later integration PR will rework error handling for unsupported input -- there currently isn't a way to explicitly reject a regex because of unsupported interactions between captures and other features -- so it should be easier to address later on.

src/libre/parser.act
silentbicycle: It says 
```
		/*
		 * Numbering for capturing groups. By convention these start from 1,
		 * and 0 represents the entire matching text.
		 */
```
which is still accurate, we are just changing how the capture group 0 for the entire matching text is represented so that we can keep a prefix/suffix match for an unanchored start/end outside of group 0. Previously we started counting from 1, because group 0 was implicit.

src/libre/re.c
silentbicycle: I didn't want to assume. Personally I find the PCRE newline/$ behavior to be really awkward. It may be a special case due to interface reasons, but it introduces a bunch of other special cases with surprising consequences.

If you want to try, setting the flag there and making changes in the native parser similar to PCRE's (https://github.com/katef/libfsm/blob/sv/updates-to-pcre-tests/src/libre/dialect/pcre/parser.sid#L648-L650) should be sufficient, I think.

src/libre/ast_analysis.c
katef: oh! that's why it didn't show. cool!

src/libre/parser.act
katef: Hm I don't think I understand. It sounds like the `.group_id` field can now never be zero, in which case I think it'd be helpful to explain in the comment that group 0 is represented elsewhere.

silentbicycle: Does this sound clearer?

> By convention group 0 implicitly represents the entire matching text (not including any prefix/suffix ignored due to an unanchored start/end), and groups 1 and on represent capture groups in the regular expression.

I would prefer to only change the phrasing once, because each change will also involve rebuilding the parsers (which contain the comment, but otherwise not change).

katef: That makes sense, and it's true for regexps in general. But what's eating at me here is I still don't understand the possible values for the `.group_id` struct member. Can that ever be 0? That's what I'm after for the comment.

PR #389, opened 2023-01-27
Performance improvenments, mainly for `fsm_determinise`

silentbicycle:
- Switch minimise to grouped edge label iterators.
- Add `uint64_t` bitset ADT.
- Switch `adt/bitset` ADT to use `uint64_t[4]` and bit parallelism.
- Add `bm_nth_word`, use it to speed up `edge_set_hasnondeterminism`.
- Replace `internedstateset` with a simpler and considerably faster implementation.
- Avoid unnecessary priority queue bookkeeping during determinisation.
- Move priority queue alloc/free outside of per-state-set loop.

The largest gains here come from the interned state set replacement -- I realized that the code in determinisation using it always constructs the state set all at once, so we don't need the complicated bookkeeping for a state set that converges on the same result when built incrementally.

### DIFF
--- a/include/adt/internedstateset.h
+++ b/include/adt/internedstateset.h
@@ -7,7 +7,10 @@
 #ifndef INTERNEDSTATESET_H
 #define INTERNEDSTATESET_H
 
+#include <stdio.h>
 #include <stdint.h>
+#include <stdbool.h>
+#include <fsm/fsm.h>
 
 struct fsm_alloc;
 struct state_set;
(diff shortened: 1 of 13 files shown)


silentbicycle:
Re-running job that failed due to a race condition in the build system (see https://github.com/katef/kmkf/issues/14).

silentbicycle:
Rerunning the job that failed because of a build race condition, again.

src/adt/edgeset.c
```diff
@@ -364,41 +364,30 @@ edge_set_contains(const struct edge_set *set, unsigned char symbol)
 int
 edge_set_hasnondeterminism(const struct edge_set *set, struct bm *bm)
 {
-	size_t i;
-
+	size_t i, w_i;
 	assert(bm != NULL);
 
+	dump_edge_set(set);
+
```
katef, line 371: `dump_edge_set()` is a no-op when logging is disabled

src/adt/edgeset.c
```diff
@@ -364,41 +364,30 @@ edge_set_contains(const struct edge_set *set, unsigned char symbol)
...
-			continue; /* no element */
-		}
+			uint64_t *bmw = bm_nth_word(bm, w_i);
 
-		if (bm_get(bm, set->b[i].symbol)) {
-			return 1;
+			if (cur & *bmw) {
+				return 1;
+			}
+			*bmw |= cur;
```
katef, line 389: introducing `bm_nth_word()` for this is neat

include/adt/internedstateset.h
```diff
@@ -22,61 +25,37 @@ struct interned_state_set_pool;
...
+ * set of states, reuse the existing identifier.
  *
- * Calling this also informs the pool that there is a live reference
- * to the state_set outside the pool, so freeing the pool will no
- * longer free the state_set. */
+ * Note: states[] must be sorted and must not contain duplicates. */
+bool
+interned_state_set_intern_set(struct interned_state_set_pool *pool,
+	size_t state_count, const fsm_state_t *states,
+	interned_state_set_id *result);
```
katef, line 48: i really like this interface

src/libfsm/determinise.c
```diff
@@ -1025,6 +984,33 @@ analyze_closures__collect(struct analyze_closures_env *env)
 		struct ac_iter *iter = &env->iters[next_i];
 		assert(iter->info.to != AC_NO_STATE);
 
+		/* If we are about to put the current iterator into the
+		 * priority queue only to pop it right back out again,
+		 * note what the next state is on the next iterator in
+		 * the queue and resume the current iterator as long as
+		 * we can. This saves a lot of time spent on pointless
+		 * queue bookkeeping. */
```
katef, line 992: I appreciate this comment!

src/adt/internedstateset.c
```diff
@@ -9,1235 +9,522 @@
...
- * order the IDs are added, which allows constant-time testing for
- * equality. Partially overlapping sets will be built using shared
- * subtrees when possible, which can save quite a lot of memory.
- *
- * A hash table is used to intern individual nodes, and reference
- * counting tracks when nodes can be returned to the freelist and
- * reused. There can be a lot of short-lived nodes constructed during
- * subtree rebalancing. Reference counting keeps the working set small,
- * but generational tracing collection may also be worth exploring at
- * some point. */
```
katef, line 35: I think an equivalent orientation comment would still be helpful, even if building the iss incrementally no longer applies

katef (approved):

PR #390, opened 2023-01-31
Detect crashing/asserting tests as failures.

silentbicycle:
Previously, tests that exited unexpectedly did not always produce output, so build targets that depended on them to produce `tests/*/resN` with "PASS" or "FAIL" never got created, and the error output from re, fsm, etc. got buried in several pages of output from make.

I think this is correct, and local testing by intentionally making `pcre-anchor` tests assert has been flagged properly, but I may be misunderstanding the test setup for some subdirectories. In all cases it outputs the `FAIL` file if the earlier stage program exits unexpectedly before producing output to check later.

### DIFF
--- a/tests/complement/Makefile
+++ b/tests/complement/Makefile
@@ -10,7 +10,7 @@ FSM=${BUILD}/bin/fsm
 
 ${TEST_OUTDIR.tests/complement}/got${n}.fsm: ${TEST_SRCDIR.tests/complement}/in${n}.fsm
 	${FSM} -p -t complement ${.ALLSRC:M*.fsm} \
-	> $@
+	> $@ || echo FAIL > ${TEST_OUTDIR.tests/complement}/res${n}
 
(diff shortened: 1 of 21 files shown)


katef (approved):
Oh! Right :D

tests/complement/Makefile
```diff
@@ -10,7 +10,7 @@ FSM=${BUILD}/bin/fsm
 
 ${TEST_OUTDIR.tests/complement}/got${n}.fsm: ${TEST_SRCDIR.tests/complement}/in${n}.fsm
 	${FSM} -p -t complement ${.ALLSRC:M*.fsm} \
-	> $@
+	> $@ || echo FAIL > ${TEST_OUTDIR.tests/complement}/res${n}
```
katef, line 13: I think we'd still want to set `$?` non-zero here (the echo will exit successfully)

silentbicycle:
Nevermind, my build script wrapper is calling `bmake` with `-k`, without that it still fails as expected.

PR #392, opened 2023-03-16
Improve determinisation performance with pairwise analysis cache

silentbicycle:
This replaces the transitive closure analysis in determinisation (essentially, everything currently within `analyze_closures_for_iss`) with a faster approach.

The previous analysis found the overall result for the entire relation of (all states in the interned state set) x (all labels appearing on outgoing edges) => (all destination states), but this tends to have a great deal of redundant work in larger FSMs, especially ones produced by combining partially overlapping FSMs.

This implementation instead analyses single states and pairs of cached analysis results. Since each cached result's labels have no overlap (each label leads to at most one state set), this reduces both the input sizes per analysis run and also the total amount of label overlap that needs to be handled at once, and then the cache eliminates redundant label analysis between duplicated subsets of state sets.

This significantly speeds up determinisation for NFAs with lots of nondeterministic edges, without negatively impacting the runtime for less nondeterministic inputs. It increases memory usage somewhat, since results that were previously derived over and over are stored, but this should be an acceptable tradeoff.

On the current branch this can lead to a slightly slower result (due to greater pressure on minimisation), but in combination with a separate incoming PR that improves minimisation the overall result should be much faster.

Also, a couple misc. bugfixes that I've had on a working branch for too long:

- `fsm_move` should set `dst->hasstart` rather than leaving it uninitialized. I haven't seen a case where this breaks anything, but still.
- Add `include/adt/hash.h` and use that in the new code.

### DIFF
--- a/include/adt/hash.h
+++ b/include/adt/hash.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2021 Scott Vokes
+ *
+ * See LICENCE for the full copyright terms.
+ */
+
+#ifndef FSM_HASH_H
+#define FSM_HASH_H
+
+#include <stdint.h>
+
+#include "common/check.h"
+
+/* 32 and 64-bit approximations of the golden ratio. */
+#define FSM_PHI_32 0x9e3779b9UL
+#define FSM_PHI_64 (uint64_t)0x9e3779b97f4a7c15UL
+
+/* A suitable hash function for individual sequentially allocated
+ * identifiers. See Knuth 6.4, Fibonacci hashing. */
+
+SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+static __inline__ uint64_t
+fsm_hash_id(unsigned id)
+{
+	return FSM_PHI_64 * (uint64_t)(id + (unsigned)1);
+}
+
+/* FNV-1a hash function, 32 and 64 bit versions. This is in the public
+ * domain. For details, see:
+ *
+ *     http://www.isthe.com/chongo/tech/comp/fnv/index.html
+ */
+
+SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+static __inline__ uint32_t
+fsm_hash_fnv1a_32(const uint8_t *buf, size_t length)
+{
+#define FNV1a_32_OFFSET_BASIS 	0x811c9dc5UL
+#define FNV1a_32_PRIME		0x01000193UL
+	uint32_t h = FNV1a_32_OFFSET_BASIS;
+	size_t i;
+	for (i = 0; i < length; i++) {
+		h ^= buf[i];
+		h *= FNV1a_32_PRIME;
+	}
+	return h;
+}
+
+SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+static __inline__ uint64_t
+fsm_hash_fnv1a_64(const uint8_t *buf, size_t length)
+{
+#define FNV1a_64_OFFSET_BASIS   0xcbf29ce484222325UL
+#define FNV1a_64_PRIME		0x100000001b3UL
+	uint64_t h = FNV1a_64_OFFSET_BASIS;
+	size_t i;
+	for (i = 0; i < length; i++) {
+		h ^= buf[i];
+		h *= FNV1a_64_PRIME;
+	}
+	return h;
+}
+
+#endif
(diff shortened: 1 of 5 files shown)


src/libfsm/determinise_internal.h
```diff
@@ -98,26 +103,68 @@ struct mappingstack {
 #define DEF_OUTPUT_CEIL 4
 #define DEF_DST_CEIL 4
 
+/* Mark an ID as a cached result ID (1) rather than a state ID (0). */
+#define RESULT_BIT ((fsm_state_t)1U << (8U*sizeof(fsm_state_t) - 1))
```
silentbicycle, line 107: Strictly speaking, I don't think the `RESULT_BIT` flag is necessary. Iit's redundant, because in all cases we _should_ be able to distinguish between result or state IDs by whether the pair of IDs are the same -- they can only match for caching the result from a single state, otherwise they have to differ. Using the flag gives an extra integrity check and makes some of the debugging printouts clearer, and the time spent on or-ing / masking the flag is probably inconsequential, but I'm a bit uneasy with how it clutters some of the dataflow. What do you think?

src/libfsm/determinise.c
```diff
@@ -1261,59 +1251,1014 @@ analyze_closures__analyze(struct analyze_closures_env *env)
...
+	const fsm_state_t a = *(const fsm_state_t *)pa;
+	const fsm_state_t b = *(const fsm_state_t *)pb;
+	return a < b ? -1 : a > b ? 1 : 0;
+}
+
+static void
+sort_and_dedup_dst_buf(fsm_state_t *buf, size_t *used)
+{
+	const size_t orig_used = *used;
+	qsort(buf, orig_used, sizeof(buf[0]), cmp_fsm_state_t);
```
silentbicycle, line 2020: Unfortunately this adds a `qsort` on the per-DFA-state path. It could be worth replacing that with a less general sort, since we know the input must be a set of valid state IDs `0 <= id < fsm->statecount` and any duplicates can be rejected. That's better off happening in a different PR, though.

src/libfsm/determinise.c
```diff
@@ -252,31 +256,57 @@ fsm_determinise(struct fsm *nfa)
 		fsm_move(nfa, dfa);
 	}
 
+#if EXPENSIVE_CHECKS
+	assert(fsm_all(nfa, fsm_isdfa));
```
silentbicycle, line 260: `fsm_minimise` also checks this, but it makes sense to check here since it if it fails, it could be a bug in `fsm_determinise` rather than API misuse.

src/libfsm/determinise.c
```diff
@@ -1261,59 +1250,1014 @@ analyze_closures__analyze(struct analyze_closures_env *env)
...
+		const size_t dst_count = e->to_set_count;
+		assert(e->to_set_offset + dst_count <= env->to_sets.used);
+		const fsm_state_t *dst = &env->to_sets.buf[e->to_set_offset];
+
+		interned_state_set_id iss;
+		if (!interned_state_set_intern_set(env->issp, dst_count, dst, &iss)) {
+			return 0;
+		}
+
+		if (!analyze_closures__save_output(env, labels, iss)) {
```
katef, line 1302: gcc 12 complains here:
```
src/libfsm/determinise.c:1302:22: warning: ‘analyze_closures__save_output’ reading 512 bytes from a region of size 32 [-Wstringop-overread]
 1302 |                 if (!analyze_closures__save_output(env, labels, iss)) {
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/libfsm/determinise.c:1302:22: note: referencing argument 2 of type ‘const uint64_t[64]’ {aka ‘const long unsigned int[64]’}
src/libfsm/determinise.c:2241:1: note: in a call to function ‘analyze_closures__save_output’
 2241 | analyze_closures__save_output(struct analyze_closures_env *env,
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      ```

File: include/adt/hash.h [hunk 5]
```diff
@@ -0,0 +1,64 @@
...
+
+/* FNV-1a hash function, 32 and 64 bit versions. This is in the public
+ * domain. For details, see:
+ *
+ *     http://www.isthe.com/chongo/tech/comp/fnv/index.html
+ */
+
+SUPPRESS_EXPECTED_UNSIGNED_INTEGER_OVERFLOW()
+static __inline__ uint32_t
+fsm_hash_fnv1a_32(const uint8_t *buf, size_t length)
```
katef, line 36: The `fsm_` prefix is just for symbols with extern linkage (ideally what's only in the public API); these would be just `hash_fnv1a_32()` here

silentbicycle: Changed in e68fdc5e.

PR #393, opened 2023-03-16
Speed up minimise by working with label sets, not individual labels.

silentbicycle:
When checking if a label can be used to partition ECs, also note which other labels led to the same state (and therefore EC). All of those labels are being checked at once, so they can be skipped in later iterations. This cuts down on redundant evaluation when large numbers of labels lead to the same state, which is very common.

This needs an interface change to the edge_set API, and commits us to `USE_EDGE_BITSET == 1` in `src/adt/edgeset.c`. The performance improvements here are coming via better opportunities for bit parallelism in that edge bitset implementation.

This is the PR mentioned in #392. It's probably worth merging this first, since it's smaller and easier to review.

### DIFF
--- a/src/adt/edgeset.c
+++ b/src/adt/edgeset.c
@@ -1225,6 +1225,42 @@ edge_set_find(const struct edge_set *set, unsigned char symbol,
 }
 
 int
+edge_set_check_edges(const struct edge_set *set, unsigned char label,
+    fsm_state_t *to_state, uint64_t labels[256/64])
+{
+	size_t i;
+	if (set == NULL) {
+		return 0;
+	}
+
+#if LOG_BITSET
+	fprintf(stderr, " -- edge_set_check_edges: symbol 0x%x on %p\n",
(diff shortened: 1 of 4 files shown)


src/adt/edgeset.c
```diff
@@ -1224,6 +1224,42 @@ edge_set_find(const struct edge_set *set, unsigned char symbol,
 	return 0;		/* not found */
 }
 
+int
+edge_set_check_edges(const struct edge_set *set, unsigned char label,
+    fsm_state_t *to_state, uint64_t labels[256/64])
```
katef, line 1229: I think introducing this means we're obliged to retire `USE_EDGE_BITSET`, because there's no equivalent implementation when that macro is not true

katef (approved):

src/adt/edgeset.c
katef: done in https://github.com/katef/libfsm/pull/394

PR #397, opened 2023-03-31
Add `re_is_literal()`

katef:
This is an interface to libre that walks a regex AST and if your regex would match a literal string, responds with the string and with the set of anchors, if any.

I added a `-t` interface to re(1):
```
; ./build/bin/re -r pcre -t '^a\.b\x2ec'
anchors: ^
literal: a.b.c
; ./build/bin/re -r pcre -t '^a\.b\x2ec[.]d$'
anchors: ^$
literal: a.b.c.d\n
```


src/libre/ast.c
```diff
@@ -912,3 +913,128 @@ ast_node_type_name(enum ast_expr_type t)
 	case AST_EXPR_TOMBSTONE: return "TOMBSTONE";
 	}
 }
+
+/*
+ * TODO: we could use this for AST->NFA conversion and decide to call
+ * re_strings() instead (also for a set of alts of strings)
+ *
+ * Return values:
```
silentbicycle, line 921: Is it worth making this an enum? `ast_is_literal` sounds like a function that would return bool in a >= C99 context, but if someone just checks `if (ast_is_literal(...)) { ...}` they will end up with surprises from the error case. That feels like an easy mistake to make, and since this is in the public API it will be a hassle to change later.

src/libre/ast.c
```diff
@@ -912,3 +913,128 @@ ast_node_type_name(enum ast_expr_type t)
...
+
+	if (*n == 0) {
+		*s = NULL;
+		return 1;
+	}
+
+	/*
+	 * We're using regular malloc() here, rather than the f_malloc() hooks
+	 * (libre doesn't have access to these hooks, but I think it's worth
+	 * saying that this is intentional regardless). Because this storage
```
silentbicycle, line 1018: Good to note.

src/libre/ast.c
```diff
@@ -912,3 +913,128 @@ ast_node_type_name(enum ast_expr_type t)
...
+		nodes = &e;
+		goto done;
+	}
+
+	if (e->type != AST_EXPR_CONCAT && e->type != AST_EXPR_ANCHOR) {
+		return 0;
+	}
+
+	/*
+	 * At this point we're considering either a single anchor,
```
silentbicycle, line 967: Since this (quite reasonably, imo) only accepts single anchors, unlike PCRE, is it worth adding a test for e.g. `^^^abc$$$` being rejected?

src/libre/ast.c
```diff
@@ -912,3 +913,128 @@ ast_node_type_name(enum ast_expr_type t)
...
+ *   1: Literal, *s may or may not be NULL (i.e. for an empty string), *n >= 0
+ *      and *anchor_start/_end are set.
+ */
+int
+ast_expr_is_literal(const struct ast_expr *e,
+	int *anchor_start, int *anchor_end,
+	char **s, size_t *n)
+{
+	size_t count;
+	const struct ast_expr **nodes;
```
silentbicycle, line 933: There's a warning about mismatched const-ness.

man/re.1/re.1.xml
```diff
@@ -40,6 +40,7 @@
 	<!ENTITY y.opt "<option>-y</option>">
 	<!ENTITY z.opt "<option>-z</option>">
 	<!ENTITY s.opt "<option>-s</option>">
+	<!ENTITY t.opt "<option>-t</option>">
```
silentbicycle, line 43: Is there a mnemonic for what 't' stands for?

src/libre/ast.c
katef: Good call, I merged #399 and added a test in caa41834

src/libre/ast.c
katef: Added an emotional support const in 6df4d595

silentbicycle (approved):

PR #401, opened 2023-04-06
fsm_remove_epsilons performance improvements via grouped edge set iterator and `edge_set_add_bulk`

silentbicycle:
Replace the implementation of `fsm_remove_epsilons` with one that takes advantage of edge sets' new grouped labels. Instead of stepping through every single labeled edge on every single state, copy over edge sets from the epsilon closure in bulk.

The previous implementation had a lot of redundant overhead because it would incrementally construct the edge sets' state sets an edge at a time, which caused the state set to repeatedly `qsort` itself to maintain invariants along the way.

For example, the command `re -rpcre '^.{0,100}$'` previously took about 15.6 seconds to complete on my laptop, now it takes about 25 msec.

Also:
- change a couple predicate functions' edge set iterators to use grouping.
- switch the timer code in minimise.c to the version in `internal.h` used by everything else.

### DIFF
--- a/src/libfsm/epsilons.c
+++ b/src/libfsm/epsilons.c
@@ -53,29 +53,25 @@ static int
 carry_endids(struct fsm *fsm, struct state_set *states,
     fsm_state_t s);
 
-static void
-free_closure_sets(struct state_set *closures[FSM_SIGMA_COUNT])
-{
-	int i;
-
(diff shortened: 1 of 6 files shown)


katef (commented):
i forgot to :+1:

PR #402, opened 2023-04-09
provide more control over generated function and package names for Go

### LINKED ISSUE #360: re's go generator should give more control over package and function names
While the -e option does allow putting a prefix on the package name of the generated go,
it is a bit awkward to require each matcher to be in its own package.

An option to provide a prefix for the Match function name would be useful,
as would one to override the generated package name entirely.

dgryski:
Fixes #360

	$ re -r pcre -k pair -e Meow -E 'matchers' -pl amd64_go meow |head -4
	#include "textflag.h"

	// func MeowMatch(data []byte) int
	TEXT    ·MeowMatch(SB), NOSPLIT, $0-32

	$ re -r pcre -k pair -e Meow -E 'matchers'-pl go meow |head -4
	package matchersfsm

	func MeowMatch(data []byte) int {
		var idx = ^uint(0)

### DIFF
--- a/include/fsm/options.h
+++ b/include/fsm/options.h
@@ -58,6 +58,9 @@ struct fsm_options {
 	/* a prefix for namespacing generated identifiers. NULL if not required. */
 	const char *prefix;
 
+	/* the name of the enclosing package; NULL to use `prefix` (default). */
+	const char *package_prefix;
+
 	/* character pointer, for C code fragment output. NULL for the default. */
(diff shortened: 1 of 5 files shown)


dgryski:
Not super happy with the choice of `E` for the flag, nor do I like that it changes the default behaviour for the Go printer (even if I think it makes it better).

katef (approved):

PR #406, opened 2023-04-12
Compatibility fix for newer make

hvdijk:
Current NetBSD make adds support for setting system include directories using .SYSPATH. This support interferes with the Makefile's attempt to override the system include path by setting -m in .MAKEFLAGS. Luckily, this is easily avoided: the old -m approach effectively has no effect on newer make, and the newer .SYSPATH approach effectively has no effect on older make, so using both works with old and new make alike.


hvdijk:
With bmake from <https://www.crufty.net/help/sjg/bmake.html>, without this PR, only bmake-20230120 and older work, bmake-20230123 and newer result in
```
bmake: "/usr/share/mk/subdir.mk" line 93: Inconsistent operator for all
bmake: "/usr/share/mk/subdir.mk" line 93: Inconsistent operator for clean
egrep: warning: egrep is obsolescent; using grep -E
bmake: "/usr/share/mk/init.mk" line 90: Inconsistent operator for all
bmake: "/usr/share/mk/subdir.mk" line 102: Inconsistent operator for all
bmake: "/usr/share/mk/obj.mk" line 46: Malformed conditional (${MK_AUTO_OBJ} == "yes")
bmake: "/usr/share/mk/dpadd.mk" line 330: Malformed conditional (${.MAKE.MODE:Mmeta*} != "" && exists(${.MAKE.DEPENDFILE}))
bmake: "/usr/share/mk/prog.mk" line 140: Inconsistent operator for all
bmake: "/usr/share/mk/subdir.mk" line 102: Inconsistent operator for all
bmake: "/usr/share/mk/final.mk" line 11: Malformed conditional (${MK_STAGING} == "yes")
bmake: "/usr/share/mk/final.mk" line 18: Inconsistent operator for install
bmake: Fatal errors encountered -- cannot continue
bmake: stopped in /home/harald/libfsm
```
With this PR, the build completes without issues.

hvdijk:
The CI failure on the first attempt was specifically because `.SYSPATH` had no special effect in older bmake, which meant it being the initial target also made it the default target when no target is explicitly specified. Updated to set `.MAIN` explicitly.

katef:
Incredible. Thanks for figuring that out, I was stumped.

PR #407, opened 2023-04-13
Fix incomplete minimisation due to edge_set label collection bug

### LINKED ISSUE #404: bug: non-minimal DFA produced by fsm_minimise
Once a match is found, the result should be returned immediately without walking the rest of the string to figure out capture group 0.

silentbicycle:
Add better testing for minimisation and fix incomplete minimisation's root cause.

Fixes #404.

## Add minimisation test oracle and minimise_test_case_list.c.

Testing was previously done with `fsm -t equal`, which checks that the state machine matches the same input as the expected result, but doesn't confirm that it is minimal. Also minimised the FSM using a reference implementation and verify that the resulting state count is as expected.

- Add src/libfsm/minimise_test_oracle.c: Inefficient but easily checked implementation of minimisation. (This does not handle DFAs with captures properly.)

- tests/minimise/minimise_test_case_list.c: For a list of regex patterns, check that minimising each ends up with the same number of states as the test oracle implementation's minimised DFA.

- Add an `EXPENSIVE_CHECK` after `fsm_minimise` -- if the FSM does not have captures, minimise it using the test oracle and check that the state count matches. Depending on the input size, this can potentially be quite expensive.

- Add `fsm_capture_has_captures`, currently an internal interface.

- Add `fsm_shuffle`, which uses a linear congruential pseudorandom number generator to shuffle state IDs. This is used to check that renumbering states in a DFA does not change minimisation's result.

- Add fuzzer harness mode for fuzzing minimisation.

The fuzzer checks whether, for an arbitrary string S, if S can be successfully converted to an NFA, whether using `fsm_shuffle` to renumber the states can lead to a minimised state count inconsistent with the result from the minimisation test oracle. Syntax errors, etc. are ignored.

To use this, run with `env MODE=m`.

After integrating the `edge_set_check_edges_with_EC_mapping` fix I ran this on 8 cores for about two hours without producing any new failures.

## Fix the bug in `edge_set`'s label collection during minimisation

Use EC membership, not state id, for label partitioning check.

When minimisation checked the states' edge sets for sets of labels leading to particular destinations it was collecting labels that led to the same state, but it should have instead collected labels to any states _within the same EC_ (which is progressively refined as information propagates during minimisation). This could lead to cases where state(s) would incorrectly split out of an EC, leading to a non-minimal DFA where some states (each representing an EC set) could be combined safely and produce a smaller DFA.

A small input that exercised this bug is "ab*c", which produced a correct but non-minimal DFA with 4 states instead of 3. This was added to `tests/minimise/minimise_test_case_list.c` as a regression.

- Rename `edge_set_check_edges` to `edge_set_check_edges_with_EC_mapping` and note that it's specific to minimisation's use case. Pass in the current EC map from minimisation and use it to determine which states should be considered equivalent to the label's first destination state.

- Also, update a variable name in a normally compiled-out logging statement and change `checked_labels`'s length to `[256/64]` to be consistent with its definition elsewhere.

## Other misc. changes

- `fsm_consolidate`'s src argument should be `const`.

- reachable.c: Removed unused variable. (Address warning.)


src/libfsm/minimise.c
```diff
@@ -256,7 +278,7 @@ build_minimised_mapping(const struct fsm *fsm,
 				}
 			}
 
-			uint64_t checked_labels[4] = {0};
+			uint64_t checked_labels[256/64] = {0};
```
katef, line 281: what's happening here?

src/libfsm/minimise_test_oracle.c
```diff
@@ -0,0 +1,279 @@
...
+	 * distinguishable). Repeat until no furhter progress can be
+	 * made. Any pairs of states still unmarked can be combined into
+	 * a single state, because no possible input leads them to
+	 * observably distinguishable results.
+	 *
+	 * This method scales poorly as the number of states increases
+	 * (since it's comparing everything against everything else over
+	 * and over, for every label present), but it's straightforward
+	 * to implement and to check its results by hand, so it works
+	 * well as a test oracle. */
```
katef, line 102: I really appreciate this comment. Also I'm amazed that you can just like, write out an implementation of this. It took me weeks to do the first implementation of NFA->DFA conversion!

src/libfsm/minimise_test_oracle.c
```diff
@@ -0,0 +1,279 @@
...
+	 * an extra dead state, which is non-accepting and only has
+	 * edges leading to itself.
+	 *
+	 * To start, mark all pairs of states in the table where one
+	 * state is an end state and the other is not. This indicates
+	 * that they are distinguishable by their result. Then, using
+	 * the Myhill-Nerode theorem, mark all pairs of states where
+	 * following the same label leads each to a pair of states that
+	 * have already been marked (and are thus transitively
+	 * distinguishable). Repeat until no furhter progress can be
```
katef, line 93: sp. furhter

src/libfsm/minimise.c
silentbicycle: It's called out in one of the commit logs but I didn't include it in the PR description:

>  change `checked_labels`'s length to `[256/64]` to be consistent with its definition elsewhere.

`256/64` us used in a bunch of other places for a character set of `FSM_SIGMA_COUNT` represented as a `uint64_t` bitset, and since you caught that typo before of `256/4` I wanted to make sure that was consistent.

katef: Oh! I actually misread this, that makes even more sense. I thought you were changing the array size here, but you're not.

src/libfsm/minimise_test_oracle.c
silentbicycle: To be fair, I wrote that in somewhat modified form about three years ago (https://github.com/katef/libfsm/blob/main/theft/fuzz_minimise.c#L306) -- this was adapting an existing thing, which was working with different input/output structures and not using ADTs added later on, rather than writing it from scratch. Looking at the commit logs, it seems like that took a couple hours initially.

src/libfsm/minimise_test_oracle.c
silentbicycle: Fixed in a1fd4801.

silentbicycle:
The CI issue was because the last definition in the file passed to `objcopy --keep-global-symbols` was getting missed by gcc at link-time, but only when using `-fsanitize=address`.

katef (approved):

PR #410, opened 2023-04-20
Add `fsm_print_vmc` mode to fuzzer.

silentbicycle:
While on an older branch I got some warnings for uninitialized memory in src/libfsm/vm/ir.c's `initial_translate_state` (for `st->strategy`). The issue seems to have been fixed already on `main`, but I added a fuzz harness to search for inputs that triggered `abort()`/assert.

I ran it for about an hour without finding any, but it's probably worth keeping the fuzzer code.

### DIFF
--- a/fuzz/target.c
+++ b/fuzz/target.c
@@ -27,6 +27,7 @@
 enum run_mode {
 	MODE_DEFAULT,
 	MODE_SHUFFLE_MINIMISE,
+	MODE_PRINT_VMC,
 };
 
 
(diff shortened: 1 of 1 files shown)


silentbicycle:
Rerunning CI, unrelated failures due to network errors in the documentation jobs.

fuzz/target.c
```diff
@@ -260,6 +261,49 @@ shuffle_minimise(const char *pattern)
...
+		if (min) {
+			if (!fsm_minimise(fsm)) {
+				assert(!"min failure");
+				return EXIT_FAILURE;
+			}
+		}
+	}
+
+	/* see if this triggers any asserts */
+	fsm_print_vmc(f, fsm);
```
katef, line 301: I think we should call all fsm_print interfaces here. This fuzzer isn't really specific to the VM-based outputs, either. Let's run everything we have.

silentbicycle: Doing so immediately revealed that several of them assert or `exit(EXIT_FAILURE)` if called with an unsupported IO mode. Should they just return instead?

silentbicycle: Okay. I can commit calling all of them that but not my local changes to make them not exit/assert, then.

PR #412, opened 2023-04-20
Add Go support to retest

dgryski:


dgryski:
Because we cant unload Go shared libraries, this leaks memory causing the PCRE test suite to fail.  We need another non-dlopen-based solution.  Back to a draft for now, but I will file another PR with the bug this found.

src/retest/runner.c
```diff
@@ -20,17 +22,26 @@ runner_init_compiled(struct fsm *fsm, struct fsm_runner *r, enum implementation
...
 	const char *cc, *cflags, *as, *asflags;
+	int src_suffix_len = 0;
 	int fd_src, fd_so, fd_o;
 	FILE *f = NULL;
 	void *h = NULL;
 
-	fd_src = mkstemp(tmp_src);
+	if (impl == IMPL_GO) {
+		/* The Go compiler needs an extension on tmp_src so it knows it's a file not a package. */
+		strcat(tmp_src, ".go");
```
moon-chilled, line 40: 10/10 would pet again

dgryski:
New version of this PR now disassembles the Go and assembly output, converts it to AT&T, assembles it, and calls it via `dlopen` with small trampoline.

dgryski:
I don't understand why these failures aren't deterministic.

katef (approved):
Thanks!

PR #416, opened 2023-04-24
fuzzer: Call all `fsm_print_*` functions in each io mode when `MODE=p`.

silentbicycle:
Continuation of #410, which was merged before this could be added (as requested).

This was previously `v`, for just `fsm_print_vmc`.

This currently fails pretty quickly, because several of the print functions print something like "unsupported IO mode" to stderr and `exit(EXIT_FAILURE)` rather than reporting an error to the callers.

### DIFF
--- a/fuzz/target.c
+++ b/fuzz/target.c
@@ -27,7 +27,7 @@
 enum run_mode {
 	MODE_DEFAULT,
 	MODE_SHUFFLE_MINIMISE,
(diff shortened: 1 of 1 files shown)


katef (approved):

katef:
I'm merging this now, I/O stuff is coming shortly, and having this in main makes that a little easier for me.

PR #418, opened 2023-04-25
Some terminology

katef:
Two small things here:
- I Removed the term "Glushkovise" where we just mean "epsilon-free".
    I'm not comfortable with using the term Glushkov NFA, because that implies using Glushkov's construction, which gives a particular NFA structure from parsing a regex. Not having epislons isn't enough to satisfy that, and I think it's confusing to mix the two things.
- Naming about errors; the `X` means "expected". And unsupported operators are not syntax errors


silentbicycle (approved):
Removing `-G` is great, because that will better suit `fsm_generate_matches`.

PR #419, opened 2023-04-25
Add `fsm_generate_matches` to cli

silentbicycle:
This adds a `-G <max_length>` option to `re` and `fsm`.

Also, add a default / basic use case callback, `fsm_generate_cb_printf`, which is what re and fsm use.

### DIFF
--- a/src/libfsm/gen.c
+++ b/src/libfsm/gen.c
@@ -16,6 +16,9 @@
 #include <adt/edgeset.h>
 #include <adt/stateset.h>
 
+#include <stdio.h>
+#include <print/esc.h>
+
 #include <assert.h>
 
 #include <ctype.h>
@@ -146,6 +149,28 @@ fsm_generate_matches(struct fsm *fsm, size_t max_length,
 }
 
 enum fsm_generate_matches_cb_res
+fsm_generate_cb_printf_escaped(const struct fsm *fsm,
+	size_t depth, size_t match_count, size_t steps,
+	const char *input, size_t input_length,
+	fsm_state_t end_state, void *opaque)
+{
+	(void)fsm;
+	(void)end_state;
+	(void)depth;
+	(void)match_count;
+	(void)steps;
+
+	assert(opaque != NULL);
+	const struct fsm_options *opt = opaque;
+
+	for (size_t i = 0; i < input_length; i++) {
+		c_escputc_str(stdout, opt, input[i]);
+	}
+	printf("\n");
+	return FSM_GENERATE_MATCHES_CB_RES_CONTINUE;
+}
+
+enum fsm_generate_matches_cb_res
 fsm_generate_cb_printf(const struct fsm *fsm,
 	size_t depth, size_t match_count, size_t steps,
 	const char *input, size_t input_length,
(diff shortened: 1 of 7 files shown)


src/re/main.c
```diff
@@ -1166,6 +1182,11 @@ main(int argc, char *argv[])
 		return 0;
 	}
 
+	if (generate_bounds > 0) {
+		int r = fsm_generate_matches(fsm, generate_bounds, fsm_generate_cb_printf, NULL);
+		return r;
```
katef, line 1187: why the variable?

src/libfsm/gen.c
```diff
@@ -145,6 +145,23 @@ fsm_generate_matches(struct fsm *fsm, size_t max_length,
...
+	fsm_state_t end_state, void *opaque)
+{
+	(void)fsm;
+	(void)input_length;
+	(void)end_state;
+	(void)opaque;
+	(void)depth;
+	(void)match_count;
+	(void)steps;
+	printf("%s\n", input);
```
katef, line 161: We're defining a line-delimited protocol with the `\n` here, so the output isn't binary. Let's escape non-printable characters (use `c_escputc_str()` for that). And the CLI tools need to heed `-X` for "always_hex" output, so I suppose that has to be an additional callback.

katef (changes requested):
github insists i type something here

src/re/main.c
silentbicycle: Oh, because I was initially freeing the `fsm` before exiting, then realized that the `atexit` handler was already doing it. I'll remove the unnecessary variable.

silentbicycle: Removed in 3ae4c9be.

src/libfsm/gen.c
silentbicycle: Changed in 8219d433.

Shoulds there a way to fully _disable_ escaping in the CLI?

src/re/main.c
katef: aha! thanks

src/libfsm/gen.c
katef: No - I don't see how we ever could disable escaping, because we're printing meaningful newlines and such, so this is always a kind of protocol

katef: That is, i don't think an unescaped `fsm_generate_cb_printf()` ever makes sense

silentbicycle: Okay. I didn't want escaping when comparing results with PCRE, but that wasn't using the CLI interface at all, and as you said that's inherently a line-oriented protocol.

silentbicycle:
CI failed because of https://github.com/katef/kmkf/issues/14 , rerunning.

silentbicycle:
CI failed again, because of katef/kmkf#14 , rerunning again.

src/libfsm/gen.c
katef: Oh I see, we don't need an additional callback because `c_escputc_str()` takes an `opt`. That's fab, thank you!

katef (approved):

katef:
Sample output:
```
; ./build/bin/re -r pcre -G 5 '^a[0-9]*b?c'
abc
abc!
abc!!
ac
ac!
ac!!
ac!!!
a0bc
a0bc!
a0c
a0c!
a0c!!
a00c
a00c!
```

PR #427, opened 2023-05-14
Error handling for fsm_print_*()

katef:
The crux of this is that we `fprintf()` a whole bunch and then check `ferror()`. And then if there was an error, return -1 from `fsm_print_*()`.

This allows returning -1 for other reasons, in particular ENOTSUP for unsupported IO APIs, which is what the fuzzer was hitting.

From the commit history:

> The API here is questionable. I wanted to return -1 on error to follow the model for IO, where e.g. fprintf is specified to return "a negative value" on error. So I'm returning 0 on success. This feels a little awkward, but it does mean this interface is similar to fprintf.
>
> Internally, rather than bubbling up errors from every call, I'm just calling ferror() on every return path. Also nothing in libfsm should call perror!

I also cleaned up retest's runner a bit, around the printing and compilation. We used to accidentally run `system()` twice in some places, which is harmless but cost us CI time. And I rewrote some awkward stuff about constructing strings into fixed-size arrays.

### DIFF
--- a/fuzz/target.c
+++ b/fuzz/target.c
@@ -1,3 +1,6 @@
+#include <sys/time.h>
+#include <unistd.h>
+
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdbool.h>
@@ -6,8 +9,7 @@
 #include <string.h>
 #include <inttypes.h>
 #include <ctype.h>
-#include <sys/time.h>
-#include <unistd.h>
+#include <errno.h>
 
 #include <fsm/fsm.h>
 #include <fsm/bool.h>
--- a/fuzz/target.c
+++ b/fuzz/target.c
@@ -302,23 +302,25 @@ fuzz_all_print_functions(FILE *f, const char *pattern, bool det, bool min, const
 	}
 
 	/* see if this triggers any asserts */
-	fsm_print_api(f, fsm);
-	fsm_print_awk(f, fsm);
-	fsm_print_c(f, fsm);
-	fsm_print_dot(f, fsm);
-	fsm_print_fsm(f, fsm);
-	fsm_print_ir(f, fsm);
-	fsm_print_irjson(f, fsm);
-	fsm_print_json(f, fsm);
-	fsm_print_vmc(f, fsm);
-	fsm_print_vmdot(f, fsm);
-	fsm_print_vmasm(f, fsm);
-	fsm_print_vmasm_amd64_att(f, fsm);
-	fsm_print_vmasm_amd64_nasm(f, fsm);
-	fsm_print_vmasm_amd64_go(f, fsm);
-	fsm_print_sh(f, fsm);
-	fsm_print_go(f, fsm);
-	fsm_print_rust(f, fsm);
+	int r = 0;
+	r |= fsm_print_api(f, fsm);
+	r |= fsm_print_awk(f, fsm);
+	r |= fsm_print_c(f, fsm);
+	r |= fsm_print_dot(f, fsm);
+	r |= fsm_print_fsm(f, fsm);
+	r |= fsm_print_ir(f, fsm);
+	r |= fsm_print_irjson(f, fsm);
+	r |= fsm_print_json(f, fsm);
+	r |= fsm_print_vmc(f, fsm);
+	r |= fsm_print_vmdot(f, fsm);
+	r |= fsm_print_vmasm(f, fsm);
+	r |= fsm_print_vmasm_amd64_att(f, fsm);
+	r |= fsm_print_vmasm_amd64_nasm(f, fsm);
+	r |= fsm_print_vmasm_amd64_go(f, fsm);
+	r |= fsm_print_sh(f, fsm);
+	r |= fsm_print_go(f, fsm);
+	r |= fsm_print_rust(f, fsm);
+	assert(r == 0 || errno != 0);
 
 	fsm_free(fsm);
 	return EXIT_SUCCESS;
--- a/src/libfsm/print/dot.c
+++ b/src/libfsm/print/dot.c
@@ -24,7 +24,7 @@
 #include <fsm/print.h>
 #include <fsm/options.h>
 
-static void
+static int
 singlestate(FILE *f, const struct fsm *fsm, fsm_state_t s)
 {
 	struct fsm_edge e;
@@ -59,7 +59,7 @@ singlestate(FILE *f, const struct fsm *fsm, fsm_state_t s)
 					fsm->opt->prefix != NULL ? fsm->opt->prefix : "",
 					st);
 
-				fputs("&#x3B5;", f);
+				fprintf(f, "&#x3B5;");
 
 				fprintf(f, "> ];\n");
 			}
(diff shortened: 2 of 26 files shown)


src/libfsm/print/dot.c
```diff
@@ -72,20 +72,19 @@ singlestate(FILE *f, const struct fsm *fsm, fsm_state_t s)
...
 				e.state);
 
-			dot_escputc_html(f, fsm->opt, e.symbol);
+			dot_escputc_html(f, fsm->opt, (char) e.symbol);
 
 			fprintf(f, "> ];\n");
 		}
 
-		return;
+		return 0;
```
sfstewman, line 80: Should you check `ferror` here?

src/libfsm/print/dot.c
```diff
@@ -161,9 +156,11 @@ singlestate(FILE *f, const struct fsm *fsm, fsm_state_t s)
 	if (unique != NULL) {
 		state_set_free(unique);
 	}
+
+	return 0;
```
sfstewman, line 160: Check `ferror` ?

src/libfsm/print/fsm.c
```diff
@@ -179,7 +151,7 @@ fprintf_state(FILE *f, const struct fsm *fsm, fsm_state_t s)
 
 		if (findany(fsm, s, &a)) {
 			fprintf(f, "%-2u -> %2u ?;\n", s, a);
-			return 1;
+			return 0;
```
sfstewman, line 154: Check `ferror` or the return value of `fprintf` ?

src/libfsm/print/fsm.c
```diff
@@ -234,56 +207,90 @@ fprintf_state(FILE *f, const struct fsm *fsm, fsm_state_t s)
...
+					return -1;
 				}
 			}
 
 			fprintf(f, "\n");
 		}
 	}
 
-	return 1;
+	return 0;
```
sfstewman, line 233: Check `ferror`?

src/libfsm/print/go.c
```diff
@@ -88,20 +88,24 @@ print_cond(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt)
...
 	if (opt->endleaf != NULL) {
-		opt->endleaf(f, op->ir_state->end_ids, opt->endleaf_opaque);
+		if (-1 == opt->endleaf(f, op->ir_state->end_ids, opt->endleaf_opaque)) {
+			return -1;
+		}
 	} else {
 		fprintf(f, "{\n\t\treturn %td\n\t}\n", op->ir_state - ir->states);
 	}
+
+	return 0;
```
sfstewman, line 108: Check `ferror`?

src/libfsm/print/irdot.c
```diff
@@ -161,17 +154,14 @@ print_grouprows(FILE *f, const struct fsm_options *opt,
 }
 
 static void
```
sfstewman, line 156: should this return `int`?

src/libfsm/print/irdot.c
```diff
@@ -74,19 +74,15 @@ print_state(FILE *f, unsigned to, unsigned self)
...
```
sfstewman, line 75: Should this return `int`?

src/libfsm/print/irdot.c
```diff
@@ -115,18 +111,15 @@ print_errorrows(FILE *f, const struct fsm_options *opt,
...
```
sfstewman, line 112: Should this return `int`?

src/libfsm/print/irjson.c
```diff
@@ -55,18 +55,14 @@ print_endpoint(FILE *f, const struct fsm_options *opt, unsigned char c)
 
 static void
```
sfstewman, line 56: Should this return `int`?

src/libfsm/print/irjson.c
```diff
@@ -86,14 +82,12 @@ print_ranges(FILE *f, const struct fsm_options *opt,
...
```
sfstewman, line 83: Should this return `int`?

src/libfsm/print/irjson.c
```diff
@@ -120,11 +114,10 @@ print_groups(FILE *f, const struct fsm_options *opt,
...
```
sfstewman, line 115: Should this return `int`?

src/libfsm/print/rust.c
```diff
@@ -87,20 +87,24 @@ print_cond(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt)
...
 	if (opt->endleaf != NULL) {
-		opt->endleaf(f, op->ir_state->end_ids, opt->endleaf_opaque);
+		if (-1 == opt->endleaf(f, op->ir_state->end_ids, opt->endleaf_opaque)) {
+			return -1;
+		}
 	} else {
 		fprintf(f, "return Some(%td)", op->ir_state - ir->states);
 	}
+
+	return 0;
```
sfstewman, line 107: Check `ferror`?

src/libfsm/print/sh.c
```diff
@@ -84,12 +84,10 @@ print_label(FILE *f, const struct dfavm_op_ir *op, const struct fsm_options *opt
 }
 
 static void
```
sfstewman, line 86: Should this return `int`?

src/libfsm/print/sh.c
```diff
@@ -142,10 +144,8 @@ print_branch(FILE *f, const struct dfavm_op_ir *op)
...
-print_fetch(FILE *f, const struct fsm_options *opt)
+print_fetch(FILE *f)
```
sfstewman, line 147: Should this return `int`?

sfstewman (commented):
Sorry for the late review.  I think this is a nice change!

There are a few things that I either misunderstood or weren't totally consistent through the changes:
- In many functions that call `fprintf` and return an `int` to indicate success/error, you check `ferror()` at the end, but in some you don't.  I'm not sure if this is intentional or not.  If it's intentional, it might be worth adding an explanatory comment. I highlighted this in a few places, but not all.
- I wasn't sure if functions with static linkage should always return `int` to indicate success/error.  Some do and some do not, and rely on checking `ferror` later.  I highlighted a number of places where they still return `void`.  I would suggest making this consistent.  Or if I've misunderstood the intent, I would consider an explanatory comment somewhere on when one chooses to return `int` or to rely on `ferror()` in a caller.

src/libfsm/print/dot.c
katef: It's not intentional, you're absolutely right, I missed a bunch. Thank you!

katef: Wait, no. Those are static functions; ferror is checked in the caller. That's fine, isn't it?

fuzz/target.c
```diff
@@ -302,23 +304,25 @@ fuzz_all_print_functions(FILE *f, const char *pattern, bool det, bool min, const
...
+	r |= fsm_print_vmc(f, fsm);
+	r |= fsm_print_vmdot(f, fsm);
+	r |= fsm_print_vmasm(f, fsm);
+	r |= fsm_print_vmasm_amd64_att(f, fsm);
+	r |= fsm_print_vmasm_amd64_nasm(f, fsm);
+	r |= fsm_print_vmasm_amd64_go(f, fsm);
+	r |= fsm_print_sh(f, fsm);
+	r |= fsm_print_go(f, fsm);
+	r |= fsm_print_rust(f, fsm);
+	assert(r == 0 || errno != 0);
```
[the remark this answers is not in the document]
katef: If errno isn't already zero there, I want to find out why. I'll add an assertion at the top on a new PR

src/libfsm/print/dot.c
sfstewman: That makes sense.  I found it confusing that you were checking for some errors and not others, so returning `0` doesn't mean "no error" so much as "no non-fprintf error".

You explained this in the PR description, but I somehow didn't pick up on it, and reading through the code I wasn't sure why you were checking `ferror` in some places and not others, and why there were statically linked functions that returned `int` and others that returned `void`.

I think if you add a comment or two in the code to explain all of this, it will probably help avoid confusion when someone goes through the code later.

PR #428, opened 2023-05-17
ast_compile: If errno != 0, and err->e is unset, set it to `RE_EERRNO`.

silentbicycle:
It's possible to trigger this assert in re.c's re_comp:

    assert(err == NULL || err->e != RE_ESUCCESS);

if using a very tightly limited custom allocator for libfsm.

Is this the right location to update `err->e` if unset?

### DIFF
--- a/src/libre/ast_compile.c
+++ b/src/libre/ast_compile.c
@@ -975,6 +975,10 @@ ast_compile(const struct ast *ast,
 
 error:
 
+	if (err != NULL && err->e == RE_ESUCCESS && errno != 0) {
+		err->e = RE_EERRNO;
+	}
+
 	fsm_free(fsm);
 
 	return NULL;


src/libre/ast_compile.c
```diff
@@ -975,6 +975,10 @@ ast_compile(const struct ast *ast,
 
 error:
 
+	if (err != NULL && err->e == RE_ESUCCESS && errno != 0) {
+		err->e = RE_EERRNO;
+	}
+
```
katef, line 981: Interesting. I don't think this is quite correct, because `errno` could already be set non-zero for reasons beyond our control, before entering the library. Generally I think inspecting `errno` only makes sense for a caller, because they can explicitly set it zero beforehand (or they know e.g. by a return status that it was set regardless of its previous value).

So I'm not sure what to do here. Where does the non-zero value come from?

silentbicycle: I could note whether it was still 0 when entering `ast_compile`, only updating it when set by something within that function.

The non-zero value is coming from malloc, or in this case the user callback wrapping malloc, which sets `errno = ENOMEM` when the allocation total reaches a configured limit.

I could also only specifically update it if `errno == ENOMEM` (plus the other conditions), because the specific issue here is that ensuring `err->e` is updated on every single code path where memory allocation can possibly fail would be a lot of extra work, but any of them could cause that assert to blow up later.

katef: okay, but shouldn't the allocation callback also return NULL in that situation?

silentbicycle: Yes, when the allocation callback returns `NULL` on any codepath within, that's why we end up in `ast_compile`'s error handler. There isn't anything on this codepath updating `err->e` to `EERRNO` if, for example, `fsm_new` fails to allocate memory for its structs, so then the assert blows up later.

katef: In this case I think we should remove the `errno != 0` condition, shouldn't we? because the API for the allocation callback is "return NULL and set errno to whatever you want" (and I don't want to _require_ errno is set non-zero there, that's a user choice). Then we just pass errno along as-is.

Meanwhile over in our other repos which use libfsm, we absolutely should be asserting errno is non-zero there.

PR #429, opened 2023-05-19
Run Rust codegen and fuzzing in CI

katef:
The Rust codegen has warnings for a few cases, which I haven't addressed here. I'd like to do that before I merge this PR.

I'm not sure if the way i'm running fuzzing is good. I'm building with UBSAN and ASAN (but not MSAN), and I'm essentially recreating the same work the `run_fuzzer` script does, but in CI. Breaking that up a bit allows caching a build between stages. The fuzzer itself runs for all modes (currently just `p` and `m`).

I fixed a couple of unrelated small things while I was here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on: [ push, pull_request ]
 env:
   pcre2: pcre2-10.40
   wc: wc
+  seeds: seeds
   build: build
   cvtpcre: build/test/retest
   prefix: prefix
@@ -377,6 +378,18 @@ jobs:
         path: ${{ env.build }}
         key: build-${{ matrix.make }}-${{ matrix.os }}-${{ matrix.cc }}-${{ matrix.debug }}-${{ matrix.san }}-${{ github.sha }}
 
+    # note we do the fuzzing unconditionally; each run adds to the corpus
+    - name: Cache seeds (mode ${{ matrix.mode }})
+      uses: actions/cache@v3
+      id: cache-seeds
+      with:
+        path: ${{ env.seeds }}-${{ matrix.mode }}
+        key: seeds-${{ matrix.mode }}
+
+    - name: mkdir seeds
+      if: steps.cache-seeds.outputs.cache-hit != 'true'
+      run: mkdir -p ${{ env.seeds }}-${{ matrix.mode }}
+
     - name: Get number of CPU cores
       uses: SimenB/github-actions-cpu-cores@v1
       id: cpu-cores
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -251,8 +288,16 @@ jobs:
...
 
+    # there's an unfixed intermittent makefile bug under -j for
+    # kmkf duplicate install targets, it's not interesting for libfsm's CI,
+    # so I'm retrying on error here. # github.com/katef/kmkf/issues/14
     - name: Install
-      run: ${{ matrix.make }} -r -j $((${{ steps.cpu-cores.outputs.count }} + 1)) -C ${{ env.wc }} BUILD=../${{ env.build }} ${{ matrix.debug }}=1 PKGCONF=pkg-config PREFIX=../${{ env.prefix }} NODOC=1 install
+      uses: nick-fields/retry@v2.8.3
+      with:
+        timeout_seconds: 10 # required, but not a problem for the kmkf bug
+        max_attempts: 3
```
silentbicycle, line 298: Great, a retry should help a lot here.

.github/workflows/ci.yml
```diff
@@ -186,28 +189,44 @@ jobs:
...
+        find . -name 'lexer.?' -exec touch '{}' \; # workaround for git checkout timestamps
+        find . -name 'parser.?' -exec touch '{}' \; # workaround for git checkout timestamps
+        ${{ matrix.make }} -r -j $((${{ steps.cpu-cores.outputs.count }} + 1)) -C ${{ env.wc }} BUILD=../${{ env.build }} ${{ matrix.san }}=1 ${{ matrix.debug }}=1 AUSAN=1 PKGCONF=pkg-config CC=${{ matrix.cc }} NODOC=1 mkdir
+        ${{ matrix.make }} -r -j $((${{ steps.cpu-cores.outputs.count }} + 1)) -C ${{ env.wc }} BUILD=../${{ env.build }} ${{ matrix.san }}=1 ${{ matrix.debug }}=1 AUSAN=1 PKGCONF=pkg-config CC=${{ matrix.cc }} NODOC=1 fuzz
+
   # testing different bmake dialects
   # the goal here is to excercise the build system, not the code
   # we don't care about e.g. different compilers here
+  #
+  # I'm including EXPENSIVE_CHECKS here just so we have some coverage
```
silentbicycle, line 215: This feels like a good compromise between EXPENSIVE_CHECKS making the tests very slow, but also making sure they're actually built and don't get missed as interfaces shift.

fuzz/target.c
```diff
@@ -303,6 +307,9 @@ fuzz_all_print_functions(FILE *f, const char *pattern, bool det, bool min, const
 		}
 	}
 
+	/* if errno isn't zero already, I want to know why */
+	assert(errno == 0);
```
silentbicycle, line 311: :+1:

.github/workflows/ci.yml
```diff
@@ -315,6 +360,107 @@ jobs:
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        san: [ FUZZER ]
+        os: [ ubuntu ]
+        cc: [ clang ]
+        make: [ bmake ]
+        debug: [ DEBUG, RELEASE ] # RELEASE=1 is a no-op
+        mode: [ m, p, d ]
```
silentbicycle, line 377: Are the seeds from these modes cached distinctly? Failing seeds from one mode are likely to pass in another, but that shouldn't be treated as a sign that the root cause of the failure has been addressed.

silentbicycle (approved):
While I'm not familiar with the details of the github CI workflows, this looks reasonable to me.

Some clarification would be good about whether the seeds from distinct fuzzer modes are tracked separately, or it's otherwise clear which specific modes the seed(s) failed in.

I don't see it in here, but this could set `BUILD_FOR_FUZZER` as added in #436.

.github/workflows/ci.yml
katef: yes, the mode is part of the cache key. look for `key:`

katef: this part:
```
    - name: Save seeds (mode ${{ matrix.mode }}-${{ matrix.debug }})
      uses: actions/cache/save@v3
      if: always()
      with:
        path: ${{ env.seeds }}-${{ matrix.mode }}
        key: ${{ steps.cache-seeds.outputs.cache-primary-key }}
```

PR #430, opened 2023-05-23
A simple attempt to identify runs of characters and generate memcmp()

katef:
I hope this code doesn't stick around, it's a simplistic approach and it's in the wrong place. But I want to find out if it actually helps or not.

From the code:

This catches situations like the "abc" and "xyz" in /^abc[01]xyz/, but not for runs in unanchored regexes. For those, we'd be better off adding VM instructions for sets of strings, and producing those from the various types of enum ir_strategy.

You can see these in the generated `-l vmc` code:
```
; ./build/bin/re -r pcre -k pair -pl vmc '^[01]abc[23]xyz'
#include <string.h>

int
fsm_main(const char *b, const char *e)
{
	const char *p;
	int c;

	p = b;
	if (p == e) return -1;

	c = (unsigned char) *p++;
	if (c <= '/') return -1;
	if (c > '1') return -1;
	if (e - p < 3 || 0 != memcmp(p, "abc", 3)) return -1;
	p += 3;

	if (p == e) return -1;

	c = (unsigned char) *p++;
	if (c <= '1') return -1;
	if (c > '3') return -1;
	if (e - p < 3 || 0 != memcmp(p, "xyz", 3)) return -1;
	p += 3;

	return 0x1; /* "^[01]abc[23]xyz" */
}
```

But this doesn't help for unanchored strings, we still have lots of `goto`s here:
```
; ./build/bin/re -r pcre -k pair -pl vmc 'abc'            
#include <string.h>

int
fsm_main(const char *b, const char *e)
{
	const char *p;
	int c;

	p = b;

l0:
	if (p == e) return -1;

	c = (unsigned char) *p++;
	if (c != 'a') goto l0;

l1: /* e.g. "a" */
	if (p == e) return -1;

	c = (unsigned char) *p++;
	if (c == 'a') goto l1;
	if (c != 'b') goto l0;
	if (p == e) return -1;

	c = (unsigned char) *p++;
	if (c == 'a') goto l1;
	if (c != 'c') goto l0;
	return 0x1; /* "abc" */
}
```

### DIFF
--- a/include/print/esc.h
+++ b/include/print/esc.h
@@ -36,6 +36,10 @@ int
 escputs(FILE *f, const struct fsm_options *opt, escputc *escputc,
 	const char *s);
 
+int
+escputbuf(FILE *f, const struct fsm_options *opt, escputc *escputc,
+	const char *buf, size_t len);
+
 void
 esctok(FILE *f, const char *s);
 
--- a/src/libfsm/print/vmc.c
+++ b/src/libfsm/print/vmc.c
@@ -138,6 +138,143 @@ print_fetch(FILE *f, const struct fsm_options *opt)
 	}
 }
 
+static size_t
+walk_sequence(struct dfavm_op_ir *op,
+	enum dfavm_op_end *end_bits,
+	char *buf, size_t len,
+	struct dfavm_op_ir **tail)
+{
+	size_t n;
+
+	assert(end_bits != NULL);
+	assert(buf != NULL);
+	assert(tail != NULL);
+
+	/*
+	 * Here we're looking for a sequence of:
+	 * 
+	 *   FETCH: (or fail)
+	 *   STOP: c != 'x' (or fail)
+	 * 
+	 * This catches situations like the "abc" and "xyz" in /^abc[01]xyz/,
+	 * but not for runs in unanchored regexes. For those, we'd be better
+	 * off adding VM instructions for sets of strings, and producing
+	 * those from the various types of enum ir_strategy.
+	 */
+
+	n = 0;
+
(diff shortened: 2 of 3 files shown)


silentbicycle (approved):
I don't have a deep familiarity with the vmc IR, but conceptually this makes sense to me.

For the unanchored case, could that be (when safe) represented as an explicit loop over a character array that jumps (back) to a particular state on mismatch, advanced in the input to the point of mismatch, rather than rejecting? At least with an unanchored prefix there's often a fixed happy path string and an unanchored start loop state that just consumes input until it starts to make headway on that. A string loop match would represent finding the substring (obviously), an incomplete match that advances input would represent spinning through the unanchored star loop for N characters, but it could be able to avoid emitting all the "...otherwise go back to the unanchored start loop state" cases for every other edge along the way.

katef:
> could that be (when safe) represented as an explicit loop over a character array that jumps (back) to a particular state on mismatch,

that would work, but I don't want to introduce backtracking!

silentbicycle:
I wasn't suggesting backtracking per se, rather instead of pass/fail it's pass/fail/"we're still at the state representing the unanchored start loop and possibly starting on the exact string match, again, but have consumed (and ignored) some of the input". A memcmp-like mechanism, except one that returns how far it advanced before mismatch, so we can avoid explicitly representing the "continue on <char>, everything else goes back at the unanchored start loop state" edges in the generated code. It wouldn't be backtracking because it would still consume the input, it'd just collapse down the representation of those three cases much like using memcmp collapses down the two (pass/fail).

PR #432, opened 2023-06-03
Emit inline strncmp() for str IO vmc codegen

katef:
This is equivalent to #430 but with `strncmp()` for `FSM_IO_STR`.

Coverage for the string API is poor. I don't want to try to run the pcre suite for that, because `FSM_IO_STR` has no provision for binary data, being a \0-terminated string. So I've extended how we run retest for a few handwritten test cases instead.

The generated code looks like this:
```
; ./build/bin/re -k str -l vmc '^abc[xy]'
int
fsm_main(const char *s)
{
	const char *p;
	int c;

	p = s;
	if (0 != strncmp(p, "abc", 3)) return -1;
	p += 3;

	if (c = (unsigned char) *p++, c == '\0') return -1;
	if (c <= 'w') return -1;
	if (c > 'y') return -1;
	return 0x1; /* "^abc[xy]" */
}
```

### DIFF
--- a/src/libfsm/print/vmc.c
+++ b/src/libfsm/print/vmc.c
@@ -392,6 +392,19 @@ fsm_print_cfrag(FILE *f, const struct ir *ir, const struct fsm_options *opt,
 				fprintf(f, "p += %zu;\n", n);
 
 				op = tail;
+			} else if (n > 1 && opt->io == FSM_IO_STR) {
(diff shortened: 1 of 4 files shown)


dgryski (approved):
LGTM

june-fastly (approved):

PR #433, opened 2023-06-06
Escape ? as a way to avoid emitting trigraphs.

katef:
This is applicable for the inline memcmp()/strncmp() calls added recently for the vmc codegen.

Before:
```
; ./build/bin/re -k str -pl vmc '^abc\?\?-$'
int
fsm_main(const char *s)
{
	const char *p;
	int c;

	p = s;
	if (0 != strncmp(p, "abc??-", 6)) return -1;
	p += 6;

	if (c = (unsigned char) *p++, c == '\0') return 0x1; /* "^abc\?\?-$" */
	(void) c;

	return -1;
}
```

After:
```
; ./build/bin/re -k str -pl vmc '^abc\?\?-$' | grep strncmp 
	if (0 != strncmp(p, "abc\077\077-", 6)) return -1;
; ./build/bin/re -k pair -pl vmc '^abc\?\?-$' | grep memcmp 
	if (e - p < 6 || 0 != memcmp(p, "abc\077\077-", 6)) return -1;
; 
```

### DIFF
--- a/src/print/c.c
+++ b/src/print/c.c
@@ -80,9 +80,11 @@ c_escputc_str(FILE *f, const struct fsm_options *opt, char c)
 	 * Escaping '/' here is a lazy way to avoid keeping state when
 	 * emitting '*', '/', since this is used to output example strings
 	 * inside comments.
+	 *
+	 * Escaping '?' is a cheap way to avoid accidentally emitting trigraphs.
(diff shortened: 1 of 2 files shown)


BinaryVixen899 (approved):
Looks good!

PR #434, opened 2023-06-08
Disallow \C

katef:
This should never have been supported, not rejecting it previously was a bug.


komiga (approved):
no units allowed :no_good_woman:

sw17ch (approved):

PR #436, opened 2023-06-13
Move internal definitions to common header + misc small bugfixes

silentbicycle:
Move some internal definitions into `include/adt/common.h` that were previously in `src/libfsm/internal.h`, since they are also used outside of `libfsm/`.

These use `#ifndef` so that they can be set via build system options.

Also:
- fix a possible NULL pointer write
- fix possible use of uninitialized memory
- add `BUILD_FOR_FUZZER` and use it in a few places.


src/libfsm/vm/ir.c
```diff
@@ -301,9 +301,9 @@ opasm_new(struct dfavm_assembler_ir *a, enum dfavm_op_instr instr, enum dfavm_op
 		op->instr = instr;
 
 		op->cmp_arg = arg;
-	}
 
-	op->ir_state = ir_state;
+		op->ir_state = ir_state;
+	}
```
katef, line 306: oh my! did you find this at runtime?

silentbicycle: scan-build found it.

src/libre/ast_analysis.c
```diff
@@ -204,6 +206,19 @@ analysis_iter(struct ast_expr *n)
...
+		 * repetition to hit memory limits. */
+#if BUILD_FOR_FUZZER
+                if ((n->u.repeat.max != AST_COUNT_UNBOUNDED && n->u.repeat.max >= 10)
+		    || (n->u.repeat.min >= 10)) {
+			fprintf(stderr, "%s: rejecting regex with {count} >= 10 (%u)\n",
+			    __func__, n->u.repeat.max);
+			return AST_ANALYSIS_ERROR_MEMORY;
+                }
+#endif
+
```
katef, line 221: Will rebase #429 for this

katef (approved):

Issue #438, opened 2023-06-28
We accidentally lost the ability to detect ambiguities


katef:
This should error:
```
; echo '/ab?c/ -> $t1; /abc/ -> $t2;' | ./build/bin/lx > /dev/null
; 
```

The error code is still there (calling `fsm_example`), but it's never reached. 

Equivalent stuff in re(1), I'm suspicious about using just one element from the array in `find_first_match_for_end_state()`, rather than checking them all:
```
    return find_match_with_id(end_id_buf[0]);
```

silentbicycle:
Do you have an example input & set of CLI flags handy for re?

katef:
An equivalent case for re, `-u` should report the ambiguities
![image](https://github.com/katef/libfsm/assets/1371085/1a41fcbb-8ef1-4c60-91cc-6cda958fbb32)

silentbicycle:
On da28efb6 (the #439 branch) this appears to be working as expected:
```
$ echo '/abc*/ -> $t1; /abc/ -> $t2;' | build/bin/lx -ldot | todot
ambiguous mappings to $t1, $t2; for example on input 'abc'
```
![out](https://github.com/katef/libfsm/assets/217280/9fdf29d9-e5b2-4c63-9b64-f74b8f8db5c9)

Am I missing something?

katef:
Oh I think `fsm_example()` is broken.

PR #439, opened 2023-07-20
Re-instate detecting ambiguous mappings in lx.

silentbicycle:
Previously the conflicts were flagged inside of carryopaque, but that was removed in 3d532eef. Now do it in its own pass, called `ast_noteconflicts`.

Add a test (`tests/lxconflict/in0.lx` and `out0.err`).

This fixes the lx side of #438.


katef:
Hm I think something is amiss here, but I can't work out what.
image


Here I think `$t1` shouldn't be matched by "b", and therefore that the example for this is wrong:
```
ambiguous mappings to $t1, $t2; for example on input 'b'
```

doing the same with re(1) (note `-b`) gives the end labels I'd expect:
image

So is lx propogating these endids too enthusiastically between states somehow?

katef:
Confirming that the above bug doesn't exist on the current `main` (which is of course the point of #438):
image

silentbicycle:
> Hm I think something is amiss here, but I can't work out what.

Your re command line has `a?b?c` whereas your lx commands `a?b?c?` (note the trailing `?`) so that's an extra difference.

Issue #386, opened 2022-10-03
Unhandled `TOK_ERROR` for incompletely parsed unsupported backtracking control flags


silentbicycle:
The PCRE dialect's lexer and parser hit an assertion with the input `(*:`:
```
$ build/bin/re -rpcre '(*:'
re: src/libre/parser.act:1108: struct ast *parse_re_pcre(re_getchar_fun *, void *, const struct fsm_options *, enum re_flags, int, struct re_err *): Assertion `!"unreached"' failed.
```

This appears to be because the lexer expects a `)` to close the backtracking control flags, but if the `'(*' .. ')' {}` in lx does not get the closing ')'` it just returns `TOK_ERROR`, and the parser does not have a catch-all error handler at the right layer -- the very first thing the generated parser's point of entry does is `return` if the current terminal is the error token. This misses the other error handlers, so the assertion that if parsing did not produce an expression, an error code is set fails shortly after.

I spent a bit trying to flag the error properly in the parser, but I'm not that familiar with `sid`. Adding a `TOK_ERROR` check in `parser.act` between `ADVANCE_LEXER;` and `DIALECT_ENTRY` catches it, but seems hacky. There also doesn't seem to be a clearly appropriate error code -- `RE_EXSUB` ("expected sub-expression") may be the closest.

silentbicycle:
Similarly, the intermediate steps for other multi-character tokens with `..`, will trigger the same assert -- `(*C` (for `(*CR)`, `(*P`, `(*pos`, `(?<`, and so on. A catch-all error hander for `TOK_ERROR` should address all of these.

PR #440, opened 2023-07-24
Integrate combinable DFA capture resolution for our PCRE dialect regexes, part 1.

silentbicycle:
This adds an extra compilation step for PCRE regexes, generating programs for a group capture resolution abstract machine. The implementation builds on the approach described in ["Regular Expression Matching: the Virtual Machine Approach"](https://swtch.com/~rsc/regexp/regexp2.html), with adaptations to reduce the runtime memory footprint, handle some edge cases the same way PCRE does, explicitly detect some other PCRE edge cases and reject them as unsupported, and continue supporting group captures as the resulting DFAs are combined with other DFAs into a single DFA that matches them all in one pass

I have fuzzed this code quite a lot, both comparing group capture results to PCRE's and comparing libfsm's handling of individual regexes' DFAs with the DFA produced my combining them. Many of the test cases in `tests/capture/capture_test_case_list.c` came from fuzzing. My working branch diverged from `main` for a while during this work, in particular the metadata associated with end states, but after integrating upstream changes I fuzzed it further. I think I integrated things properly, but it's something to be mindful of during review. Some unrelated issues discovered during fuzzing have already been posted as separate PRs.

Worst-case memory usage is proportional to the size of the capvm program's length (known at compile time) and the input length -- we record a bit for each branch taken while evaluating the regex (e.g. either greedily jumping to repeat a subexpression or non-greedily advancing in the regex), so memory usage will slowly increase as inputs are evaluated. Common path prefixes are shared between threads, and the total thread count (and divergence) is bounded by the size of the opcode program. As a special case, a long path prefix of all zero bits is collapsed down to a counter; this usually happens because of an unanchored start loop, where each zero bit represents a path continuing to advance through the input without starting the match yet.

This PR targets an integration branch because some work is not yet complete:

- Some quick cleanup of CI issues -- the submodule's build config is more strict about variables that are set but unused (because their use is in logging code that compiles away).
- A few pathological cases having to do with repetition in combination with anchors are rejected as unsupported, such as `^(()($)|x)+$`. These are probably not worth the trouble to support. It also doesn't support the `\z` operator yet.
- Some of the combinations of CLI flags for `re` aren't enabled yet, such as capture resolution for multiple files, and the `re` docs need updating for some new flags.
- We should be able to calculate likely memory usage info at regex compile time, which could be used by the caller to stack-allocate and reuse a preallocated buffer rather than using dynamic allocation at runtime, but I haven't implemented that yet. The memory usage should small in practice, but it would be nice to completely eliminate runtime dynamic allocation. The path metadata table currently grows on demand, the other data structures are either fixed-size or the same length as the opcode count. Cases where it would hit the caller's memory limit for growing the path table are likely to hit PCRE's limits much sooner.
- Code generation does not output the capture resolution abstract machine's opcodes yet. When using capture handling, libfsm's caller will need to either link a small implementation of the abstract machine or have its code generation produce a standalone implementation, but because that brings some architectural trade-offs I wanted to wait until this round has been reviewed.


silentbicycle:
I've merged the current `main` into this and fuzzed it for another 20 minutes, on 8 CPU cores.

fuzz/target.c
```diff
@@ -23,17 +24,47 @@
...
 
+/* for TRACK_TIMES and EXPENSIVE_CHECKS */
+#include "../src/libfsm/internal.h"
+
 enum run_mode {
-	MODE_DEFAULT,
+	MODE_REGEX,
+	MODE_REGEX_SINGLE_ONLY,
+	MODE_REGEX_MULTI_ONLY,
+	MODE_IDEMPOTENT_DET_MIN,
```
katef, line 59: Can you add these new modes to ci.yml too please? where currently we have:
```
mode: [ m, p, d ]
```

fuzz/target.c
```diff
@@ -143,21 +290,1069 @@ codegen(const struct fsm *fsm)
...
+
+cleanup:
+	free(captures);
+	return res;
+
+}
+
+static int
+compare_with_pcre(const char *pattern, struct fsm *fsm)
+{
```
katef, line 752: fantastic

fuzz/target.c
```diff
@@ -369,12 +1649,87 @@ harness_fuzzer_target(const uint8_t *data, size_t size)
...
+				 * particular fuzzer. */
+				return EXIT_SUCCESS;
+			}
+		}
+
+		if (c == '(') {
+			/* This triggers an "unreached" assertion in the parser.
+			 * It's already been reported (issue #386), but once the
+			 * fuzzer finds it, it will report it over and over.
+			 * Exit here so that the fuzzer considers it uninteresting. */
```
katef, line 1683: #386

fuzz/target.c
silentbicycle: Updated in 239927c3.

PR #441, opened 2023-08-30
Performance improvements for epsilon removal and determinisation

silentbicycle:
This PR improves performance for a couple code paths in epsilon removal and determinisation that become hotspots with large `{}` repeated groups:

- Add a fast path upfront for when `state_set_search` would return the position after the array (appending), since this is very common.
- Switch from linear to binary searching in `edge_set_add_bulk`.
- Avoid unnecessary calls to `qsort`, when working space would be small (also very common) build a bitset on the stack instead.

I did some other experimentation with reworking the algorithm for `epsilon_closure` to avoid populating the closure with intermediate states on the transitive path to the epsilon closure endpoints, but in the end it didn't have anywhere near as much impact as the other changes, and it makes the epsilon closure function less general-purpose, so it's probably not worth the extra complexity. For epsilon removal we only need to add states to the closures that either have labeled edges or are an end state. Filtering anything else (potentially including states being in their own epsilon closure) reduces the state set sizes for further processing. Otherwise they get filtered out in a later step.

On my laptop, building with `-O3`, the difference between `main` and this PR's `HEAD` for `time ./re_main -rpcre -C '^[ab]{0,5000}$'`:
- `main`: `real	0m41.328s`
- `HEAD`: `real	0m5.167s`

I chose that regex for benchmarking because it produces a very long chain of epsilon closures and a big edge set pileup.

### DIFF
--- a/src/adt/edgeset.c
+++ b/src/adt/edgeset.c
@@ -11,6 +11,7 @@
 #include <inttypes.h>
 
 #define LOG_BITSET 0
+#define LOG_BSEARCH 0
 
 #include "libfsm/internal.h" /* XXX: for allocating struct fsm_edge, and the edges array */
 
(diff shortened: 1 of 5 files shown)


silentbicycle:
Along with a couple unrelated USBan warnings (fixed above to reduce noise) CI found a bug -- `./re -rpcre '(?:c|d)(?:)(?:aaaa(?:)(?:bbbb)(?:bbbb(?:))(?:bbbb(?:)(?:bbbb)))'` should match "caaaaaaaabbbbbbbbbbbbbbbbbbbb" but doesn't -- so I'm investigating that.

silentbicycle:
Note: We can't use `bsearch` in `state_set_search` and `edge_set_add_bulk` because its interface doesn't match our use case -- `bsearch` returns NULL if the value isn't present, whereas we need the offset where the value would go if not present.

src/adt/edgeset.c
```diff
@@ -184,6 +185,100 @@ edge_set_advise_growth(struct edge_set **pset, const struct fsm_alloc *alloc,
...
+enum fsp_res {
+	FSP_FOUND_INSERT_POSITION,
+	FSP_FOUND_VALUE_PRESENT,
+};
+
+/* Use binary search to find the first position N where set->groups[N].to >= state,
+ * which includes the position immediately following the last entry. Return an enum
+ * which indicates whether state is already present. */
+static enum fsp_res
+find_state_position(const struct edge_set *set, fsm_state_t state, size_t *dst)
```
katef, line 197: Should we use this in `edge_set_find` and `edge_set_contains` too?

katef (approved):

silentbicycle:
The bug I mentioned above (with `(?:c|d)(?:)(?:aaaaaaaa(?:)(?:bbbbbbbb)(?:bbbbbbbb(?:))(?:bbbbbbbb(?:)(?:bbbbbbbb)))`) is fixed by the commit immediately after (a58cab2). At the time a comment seemed redundant, but noting that just in case.

PR #447, opened 2023-10-19
Add error code for unsupported PCRE cases (RE_EUNSUPPPCRE), reject one.

silentbicycle:
There's extra error codes in #440 for regexes that aren't UNSATISFIABLE per se, but depend on particular corner cases in PCRE that probably aren't worth supporting in an automata-based implementation.

Add a test case for one, tests/pcre/in48.re: `^a|$[^x]b*`

This is a tricky one to handle properly; according to PCRE it should match either "a<anything...>" OR "\n", but nothing else. The newline match is because $ is a non-input-consuming check that evaluation is either at the end of input, or at a newline immediately before the end. In this case `$[^x]b*` matches exactly one newline; it's equivalent to "$\n". This probably isn't worth supporting, but we can detect cases where a potential newline match appears after a $ and reject them as an unsupported PCRE behavior.


src/libre/ast.h
```diff
@@ -98,6 +98,7 @@ enum ast_flags {
 	AST_FLAG_ANCHORED_START  = 1 << 6,
 	AST_FLAG_ANCHORED_END    = 1 << 7,
 	AST_FLAG_END_NL          = 1 << 8,
+	AST_FLAG_MATCHES_1NEWLINE= 1 << 9,
```
katef, line 101: lol

silentbicycle: There will be even more niche start/end anchor flags coming soon. Expressing PCRE's anchoring rules statically gets complicated.

src/libre/ast_analysis.c
```diff
@@ -552,6 +552,46 @@ set_flags_subtree(struct ast_expr *n, enum ast_flags flags)
...
+		if ((n->u.range.from.type == AST_ENDPOINT_LITERAL) &&
+		    (n->u.range.to.type == AST_ENDPOINT_LITERAL)) {
+			return n->u.range.from.u.literal.c <= '\n'
+			    && n->u.range.to.u.literal.c >= '\n';
+		} else if ((n->u.range.from.type == AST_ENDPOINT_CODEPOINT) &&
+		    (n->u.range.to.type == AST_ENDPOINT_CODEPOINT)) {
+			    return n->u.range.from.u.codepoint.u <= '\n'
+				&& n->u.range.to.u.codepoint.u >= '\n';
+		} else if (n->u.range.from.type == AST_ENDPOINT_NAMED) {
+			/* TODO: unreachable? */
```
katef, line 579: hm. i'm not sure i understand why this would be unreachable. what happens for a class like `:space:`?

silentbicycle: `./re -rpcre -p '^abc[[:space:]]de$'` still comes in as `AST_ENDPOINT_LITERAL`, if I put a trace here I get:
```
RANGE TYPES [0 - 0]
-- from 0x09, to 0x0d
```

and likewise ```
$ ./re -rpcre -p '^abc[\d]de$'
RANGE TYPES [0 - 0]
-- from 0x30, to 0x39
```
looking at the parser, it should be coming in as a `NAMED_CLASS`, but something may be rewriting it along the way?

I couldn't figure out a way to reach that branch. I think at one point I locally added an assert there, so if the fuzzer found a way it'd blow up and let me know.

**Comment by silentbicycle** [comment]
I merged in a recent main and spent a bit more time trying to actually hit the possibly unreachable branch above.

File: src/libre/ast_analysis.c (hunk 2 above)
Comment (katef) [reply src/libre/ast_analysis.c]: Hm. If it is unreachable, let's keep an assert there. I _think_ it would be reached by _ranges_ which end in a class:
```
; ./build/bin/re -r native -pl ast  '^abc[x-[:space:]]de$'
/^abc[x-[:space:]]de$/:8: Unsupported operator
```
but apparently we don't support that.

silentbicycle: Added an assert in f5b13006.

katef: Thanks!

katef (approved):

PR #450, opened 2023-10-30
Reject nodes with AST flags set for re_is_literal()

### LINKED ISSUE #449: Case-insensitive regexps erroneously identified as literals
Not a literal, contentiously:
```
; ./build/bin/re -t -r pcre '(?i)abc'
anchors: 
literal: abc
```
also not a literal, but less contentious:
```
; ./build/bin/re -t -r pcre 'x(?i)abc'
anchors: 
literal: xabc
```
Note to self to include `something(?:y(?i)abc)` in the test cases, so we force this by AST walking rather than by checking the global ICASE flag.

katef:
This PR has the effect of rejecting `(?i)` (and various other pcre flags) in patterns consider literals by `re_is_literal()`.

The idea is that these aren't literals, because (being case insensitive), they have multiple valid matches, and a literal only has one way to match.

This resolves #449.

One concession here is that I'm allowing `(?n)` because we don't currently indicate this in the AST.

I'm also explicitly permitting `RE_END_NL`, because by the API contract this is the caller's responsibility.


silentbicycle (approved):
Makes sense to me.

PR #461, opened 2024-04-04
Minimisation should not combine end states with distinct sets of endids.

silentbicycle:
This was adapted from the group capture integration branch. It isn't specific to group capture.

Currently, metadata on end states will not prohibit merging, which can lead to false positives when using endids to detect which of the original regexes are matching when several regexes are compiled and then unioned into a single DFA. This adds a pass before the main minimisation loop that splits the ECs of end states with distinct sets of end IDs, which is sufficient to prevent them from merging later.

I reworked the test tests/endids/endids2_union_many_endids.c somewhat, since it was previously checking that the end states WERE merged. Unfortunately the other checks are somewhat weakened -- I removed the part checking that `endids[j] == j+1` because that ordering isn't guaranteed anymore: the patterns aren't anchored, so some examples will match more than one of them, and the endids collected can be nonconsecutive.

I added a separate test, endids10_minimise_partial_overlap.c, which checks the minimal case that when the two regexes /^abc$/ and /^ab*c$/ are combined and a distinct endid is set on each, matching "ac" "abc" "abbc" gets one or both endids as appropriate. Ideally, we'd have a property test that checks that any arbitrary set of regex patterns has the same set of inputs matching -> endids when they are run individually vs. when combined into a single FSM, determinised, and minimised.

### DIFF
--- a/src/libfsm/minimise.c
+++ b/src/libfsm/minimise.c
@@ -17,10 +17,6 @@
 #include <fsm/walk.h>
 #include <fsm/alloc.h>
 
-#if EXPENSIVE_CHECKS
-#include <fsm/print.h>
-#endif
-
 #include <adt/edgeset.h>
 #include <adt/pv.h>
 #include <adt/set.h>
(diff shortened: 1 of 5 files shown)


silentbicycle:
I fixed two things caught by the CI run:
- there were a couple functions called in the `EXPENSIVE_CHECKS` build that didn't get integrated
- the new test wasn't freeing the endid buffers

But I still need to update the minimization test oracle to prevent it from merging end states with distinct sets of endids.

katef (approved):

PR #463, opened 2024-04-26
Break a dependency on group #0 for re_is_literal()

katef:
Two small things here:
 * Allow bases=NULL for `fsm_union_array()`
 * A bugfix for `re_is_literal()` where I'd assumed the root of the AST is always group #0, but that's only the case for dialects providing group capture.
 
 Now that I'm writing that out, I actually think it makes more sense to have all dialects construct #0. hm

### DIFF
--- a/include/fsm/bool.h
+++ b/include/fsm/bool.h
@@ -44,7 +44,7 @@ fsm_union(struct fsm *a, struct fsm *b,
  * starting states and (if used) capture bases.
  *
  * bases[] is expected to have the same count as fsms[], and
- * will be initialized by the function.
+ * will be initialized by the function. May be NULL.
  *
  * On success, returns the unioned fsm and updates bases[].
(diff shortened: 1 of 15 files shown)


katef:
In [4b8276a](https://github.com/katef/libfsm/pull/463/commits/4b8276aae4ed6a8cce880e15894a749a48a15127) I did revert this idea, and now we have #0 constructed for all dialects instead.

silentbicycle (commented):
This should work, but do we need to add checks for whether the dialect actually uses group capture before doing something elsewhere because of that group 0 node in the AST? Otherwise it might cause different behavior in code generation once group capture support is merged or something.

Allowing `bases == NULL` looks good.

katef:
My intention with that is that all dialects will support group capture, at codegen time. Just that for most of them, the only available group will be #0.

PR #464, opened 2024-04-29
Add endid support to `re_strings`.

silentbicycle:
Change the `re_strings` API to optionally define an endid for each string.

- This stores the set of endids in a state_set. The IDs are the same size and it doesn't seem worth defining a distinct ADT.
- `re_strings_add_raw` and `re_strings_add_str` now take an extra argument, a pointer to an endid to set (if non-NULL).
- Add two tests, one that will produce a trie of `[abc]{3}` and the other showing that duplicated strings will collect all their endids, not just one.

### DIFF
--- a/src/libre/ac.c
+++ b/src/libre/ac.c
@@ -22,13 +22,13 @@ enum { POOL_BLOCK_SIZE = 256 };
 struct trie_state {
 	struct trie_state *children[256];
 	struct trie_state *fail;
+	/* use a state set as an endid set */
+	struct state_set *endids;
+
 	fsm_state_t st;
 	unsigned int index;
 	unsigned int output:1;
 	unsigned int have_st:1;
-
-	/* use a state set as an endid set */
-	struct state_set *endids;
 };
 
 struct trie_pool {
--- a/src/libre/ac.c
+++ b/src/libre/ac.c
@@ -16,6 +16,7 @@
 
 #include "ac.h"
 
+#define ENDID_NONE ((fsm_end_id_t)-1)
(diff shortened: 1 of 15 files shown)


src/libre/ac.c
```diff
@@ -25,6 +26,9 @@ struct trie_state {
 	unsigned int index;
 	unsigned int output:1;
 	unsigned int have_st:1;
+
+	/* use a state set as an endid set */
+	struct state_set *endids;
```
katef, line 31: could you put this above the `fsm_state_t` please? i think it'll help keep things grouped for struck packing

src/libre/ac.c
```diff
@@ -126,7 +135,7 @@ trie_create(void)
 }
 
 struct trie_graph *
-trie_add_word(struct trie_graph *g, const char *w, size_t n)
+trie_add_word(struct trie_graph *g, const char *w, size_t n, const fsm_end_id_t *endid)
```
katef, line 138: i don't think `endid` needs to be a pointer here

silentbicycle: It's a pointer because it's an optional argument and can be NULL, coming from `re_strings_add_str`. What I'm going for here is `Option<fsm_end_id_t>` in Rust terms. Instead of using a pointer it could be two versions of the same function call, or have a `ENDID_NONE` constant (probably `((fsm_end_id_t)-1)`), would you prefer one of those?

src/libre/ac.c
silentbicycle: Sure, done in 8b693385.

katef (approved):

PR #466, opened 2024-05-09
lx: Add a guard for referencing `lx->end.col`, may not be in the struct.

silentbicycle:
This field doesn't exist when lx is called with `-x pos`, so don't include references to it the generated code.


silentbicycle:
This bug was introduced in 20ed2561b.

katef (approved):

PR #470, opened 2024-05-25
Add fsm_intersect_charset(), fsm -U

katef:
Add `fsm_intersect_charset()`, a convenience to intersect an fsm against a given character set. I pulled this out of the forthcoming rx(1) tool.

Unlike `fsm_intersect()`, this doesn't automatically determinise its operands. I'm not convinced `fsm_intersect()` should do that, either. In most cases we care enough about performance to do this (or rather to avoid doing it unnecessarily) in the caller.

I've exposed this as `fsm -U <charset>`:

image

### DIFF
--- a/include/fsm/bool.h
+++ b/include/fsm/bool.h
@@ -56,6 +56,14 @@ struct fsm *
 fsm_intersect(struct fsm *a, struct fsm *b);
 
 /*
+ * A convenience to intersect against a character set, rather than
+ * a pre-existing FSM. Unlike fsm_intersect(), the FSM is required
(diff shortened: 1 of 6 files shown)


katef:
I removed group parsing as an entry point into a regex grammar in 2ede0ed4fc1680f61d1e9fd5eda32d74c2f28fd0, which was exposed as a similar cli argument for re(1). I think I'd either want to reintroduce that (somehow...), or invent a tr(1)-style syntax (perhaps as its own dialect?) just for this. But I wasn't ready to do that right now.

PR #471, opened 2024-05-29
Remove lexical position annotations for AST nodes

katef:
We had `struct ast_pos`, which carried source position information from a lexical token. This was present in a couple of AST nodes.

We also construct an AST backwardsly, from walking an FSM. It doesn't make sense (to me) to track source information all the way through the FSM just for the sake of annotating the AST there. So we had a couple of placeholder hacks with dummy line information.

I wasn't sure what I wanted to do about this (e.g. I considered pointing at this per-token position information from every AST node), until I realised we only actually need to report the position information on syntax errors, and those are only ever done from the parser actions. We never actually give them in an error from inspecting an AST after construction.

From the first commit here:

> These are used for reporting errors, and the only error here is reported from within the <count-range> action. In fact I think all errors involving lexical positions are reported within actions (i.e. during parse), because they are all syntax errors. And so I think we never need to annotate the AST with positions.

I removed these node-by-node, taking care to check the adjacent error reporting still works.

### DIFF
--- a/src/libre/ast.c
+++ b/src/libre/ast.c
@@ -153,22 +153,12 @@ ast_free(struct ast *ast)
 }
 
 struct ast_count
-ast_make_count(unsigned min, const struct ast_pos *start,
-    unsigned max, const struct ast_pos *end)
+ast_make_count(unsigned min, unsigned max)
 {
 	struct ast_count res;
 
-	memset(&res, 0x00, sizeof res);
-
 	res.min = min;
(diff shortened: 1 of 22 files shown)


katef:
While I'm here I also cleaned up some of the macros around parser.act, and removed the dialect-dependent state that was kept at a wider scope than it needed to be.

silentbicycle (approved):

PR #474, opened 2024-05-30
Add fsm_new_statealloc()

katef:
This allows pre-allocation where the number of states is known empirically by the caller. I added calls in a couple of places where over-allocating has been bugging me.

Based on #472, I'll rebase before this goes in.

### DIFF
--- a/src/libfsm/fsm.c
+++ b/src/libfsm/fsm.c
@@ -44,10 +44,14 @@ free_contents(struct fsm *fsm)
 }
 
 struct fsm *
-fsm_new(const struct fsm_options *opt)
+fsm_new_statealloc(const struct fsm_options *opt, size_t statealloc)
 {
 	static const struct fsm_options defaults;
 	struct fsm *new, f;
+
+	if (statealloc == 0) {
+		return fsm_new(opt);
+	}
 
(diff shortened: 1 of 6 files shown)


src/libfsm/fsm.c
```diff
@@ -41,11 +44,15 @@ free_contents(struct fsm *fsm)
 }
 
 struct fsm *
-fsm_new(const struct fsm_options *opt)
+fsm_new_statealloc(const struct fsm_options *opt, size_t statealloc)
 {
 	static const struct fsm_options defaults;
 	struct fsm *new, f;
 
+	if (statealloc == 0) {
```
silentbicycle, line 52: OK, so a hint of 0 is explicitly using the default.

silentbicycle (commented):
This looks good to me, and I was thinking about doing the same thing.

Is there anything that depends on `fsm->statealloc` being a power of 2? `fsm_vacuum` repeatedly halves the size, when it's worth doing, but that should still be fine even if it's not.

It sounds like you're about to rebase to include other recent PRs, so I'll approve once that's set.

silentbicycle (approved):

PR #475, opened 2024-05-31
Add `void* opaque` to vmc codegen too, disable unused warning.

silentbicycle:
This is a follow-up to #465:

- Also add the extra `void *opaque` to vmc's codegen.

- Add a `(void)` cast to suppress warnings if the extra opaque void pointer isn't used by the generated code.


katef (approved):

katef:
I'm having second thoughts about this, I don't like the casts for two reasons:
* They break compatibility to C89-compliant codegen
* I'd much rather not have a parameter when we don't use it, than to emit it and mark it unused

I'm going to revert this for the moment.

If the `void *opaque` parameters are useful for the generated code from the leaf/endleaf callbacks, I think perhaps we can have a way to optionally emit a parameter just for those. Or perhaps add a way to emit whatever parameters those callbacks might use, then they wouldn't have to be `void *`.

PR #482, opened 2024-07-01
Sort endids written out by fsm_endid_get()

katef:
This is for caller convenience, I keep calling qsort for various reasons whenever I call this function, for example to use memcmp on ids.

But the final motivation here was that I wanted to memcmp a const-qualified pointer to an id set stored elsewhere, and I can't do that without allocating for a local copy. And that seems silly when we can just qsort here.


silentbicycle (approved):
Sure.

This shouldn't be an issue from a performance standpoint. We compare endids for minimisation, but that's done in one pass rather than every iteration, and we could switch to a less general sort (as we have elsewhere) if it still manages to pop up in profiling.

PR #485, opened 2024-07-26
Introduce ambig mode

katef:
This PR introduces `(struct fsm_options).ambig`:
```
    AMBIG_NONE     = 0, /* default */
    AMBIG_ERROR    = 1 << 0,
    AMBIG_EARLIEST = 1 << 1,
    AMBIG_MULTIPLE = 1 << 2,
```
and provides codegen for single, none, and multiple endids on accepting states.
For most languages this means returning a boolean to indicate success/failure,
independently of whether any endids are present.

There's an API change as a side effect, `fsm_print()` is now a single interface to print (previously this was one function per language). I've also separated the `struct fsm_options` and alloc hooks from various parts; in particular `struct fsm_options` is only passed to the print routines, and is no longer attached to `struct fsm`.

Other than the API changes, and perhaps quietly fixing some bugs along the way, this PR attempts to keep existing functionality unchanged.


examples/bm/libfsm.c
```diff
@@ -80,7 +80,7 @@ main(int argc, char *argv[])
 	printf("#include <time.h>\n");
 	printf("\n");
 
-	fsm_print_c(stdout, fsm);
+	fsm_print(stdout, fsm, FSM_PRINT_C);
```
silentbicycle, line 83: Nice.

include/fsm/print.h
```diff
@@ -10,6 +10,11 @@
 struct fsm;
 struct fsm_options;
 
+/* a convenience for debugging */
+void fsm_dump(FILE *f, const struct fsm *fsm,
```
silentbicycle, line 14: :+1:

silentbicycle (approved):
There are a lot of line changes here due to pervasive interface changes, and I didn't look all at the language-specific formatting code that closely (in particular, I'm not that familiar with go syntax), but the overall changes make sense to me.

PR #487, opened 2024-08-17
Document -F <flags> and -k <io>

katef:
Documenting `-F <flags>` and `-k <io>`, which were for some reason missing. Probably because the stock docbook rendering for manpages doesn't render tables, so the output for the table here is just missing. Great stuff. I definitely want to add my own mandoc rendering at some point, but not today.


katef (commented):
Good job Kate

PR #489, opened 2024-08-23
In `closure_free()`, use the FSM allocator instead of `free()`.

sw17ch:
`closure_free()` should use the free implementation provided by the FSM's allocator.

While attempting to make bindings to the libfsm API for a Rust project, I came across a case where my allocator was giving me the same pointer twice. I added instrumentation that places guard regions on either side of the allocation, and with this, I was able to observe `closure_free()` attempting to free the internal pointer between the guards rather than the pointer to the start of the allocation.

With this change in place, I was able to run my test program to completion.

### DIFF
--- a/src/fsm/main.c
+++ b/src/fsm/main.c
@@ -666,7 +666,7 @@ main(int argc, char *argv[])
 				printf("\n");
 			}
 
-			closure_free(closures, fsm->statecount);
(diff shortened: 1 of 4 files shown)


katef (approved):

katef:
Thanks!

Issue #478, opened 2024-06-07
Behavior of `-G length` inconsistent, moreover returns error code.


pierreganty:
Hi,

When asking `re` for an example of length up to 20 then none is returned.
If, for the same regular expression, I ask `re` to return an example of length 21 then it returns two examples, the first of which has length 20.

Moreover, in both calls `re` returns `1` which is an error code according to the man page. I am not sure why, though.

```
: ~/.local/libfsm/bin ; ./re  -b -G20 '([A-Z0-9]+);([^;]+);([^;]+);([0-9]+);([^;]+);([^;]*);([0-9]*);([0-9]*);([-0-9/]*);([YN]);([^;]*);([^;]*);([^;]*);([^;]*);([^;]*)'; echo $?
1
: ~/.local/libfsm/bin ; ./re  -b -G21 '([A-Z0-9]+);([^;]+);([^;]+);([0-9]+);([^;]+);([^;]*);([0-9]*);([0-9]*);([-0-9/]*);([YN]);([^;]*);([^;]*);([^;]*);([^;]*);([^;]*)'; echo $?
0;!;!;0;!;;;;;N;;;;;
0;!;!;0;!;;;;;N;;;;;!
1
```

katef:
Finally getting to this, sorry it's taken so long!

I think this is a more compelling example:
```
; ./build/bin/re -r literal -G 3 abc
; ./build/bin/re -r literal -G 4 abc
abc
; 
```

katef:
Also a good case:
```
; ./build/bin/re -r literal -G 1 '' 

; ./build/bin/re -r literal -G 1 'x'
; 
```

pierreganty:
Nice! Thank you.

PR #490, opened 2024-08-24
A few small bugfixes

katef:
- Reported by @pierreganty: #478, off-by-one for `re -G <max-length>` and also in the same ticket, wrong exit status for `re -G` 
- Suggested by @sw17ch: `-C opt-level=3` does help for performance of rust generated code, and I'd also hoped it'd improve time in CI. But sadly no, that's still dominated by the rustc compile time.
- Spotted by @sw17ch: wrong return type for AMBIG_ERROR/AMBIG_EARLIEST for rust generated code

### DIFF
--- a/src/libfsm/gen.c
+++ b/src/libfsm/gen.c
@@ -193,7 +193,7 @@ gen_init_outer(struct fsm *fsm, size_t max_length,
     fsm_generate_matches_cb *cb, void *opaque,
     bool randomized, unsigned seed)
 {
-	int res = 0;
+	int res = false;
 	if (fsm == NULL || cb == NULL || max_length == 0) {
 		return false;
 	}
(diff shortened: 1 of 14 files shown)


src/libfsm/gen.c
```diff
@@ -562,8 +568,8 @@ sfs_step_edges(struct gen_ctx *ctx, struct gen_stack_frame *sf)
 		sf->u.step_edges.initialized = true;
 	}
 
-	if (ctx->buf_used + ctx->sed[sf->s_id] >= ctx->max_length) {
-		LOG(2, "PRUNING due to max length: used:%zu + sed[%d]:%u >= max_length:%zu\n",
+	if (ctx->buf_used + ctx->sed[sf->s_id] > ctx->max_length) {
+		LOG(2, "PRUNING due to max length: used:%zu + sed[%d]:%u > max_length:%zu\n",
```
katef, line 572: @silentbicycle requesting your review for this please, I was unsure about this.

silentbicycle: I think `>` is correct here -- this means "if the buffer used so far plus what we'd need for the shortest distance to an end state exceeds the available buffer". `re -G 5 '^abcde'` should produce "abcde" as a valid 5-character match (which is rejected with `>=`), not counting the trailing `\0`, but reject `re -G 5 '^abcdef'`.

silentbicycle (approved):

PR #492, opened 2024-09-12
Add fsm_detect_required_characters

silentbicycle:
This inspects the DFA to determine which characters must appear in any matching input.

Since this can be very slow for large/complex DFAs there is an optional step limit.

It walks paths between start and end states, noting character constraints from unavoidable edges with only a single label, and intersects requirements over all the paths found. It terminates early if there no constraints remaining, because further processing would not learn anything new.


src/libfsm/detect_required.c
```diff
@@ -0,0 +1,275 @@
...
+	assert(any);
+}
+
+/* Walk a DFA and attempt to detect which characters must appear in any input to match.
+ * This finds the intersection of characters required on any start->end paths (tracking
+ * edges with only one label that must be followed by all matches), so it can take
+ * prohibitively long for large/complex DFAs. */
+enum fsm_detect_required_characters_res
+fsm_detect_required_characters(const struct fsm *dfa, size_t step_limit,
+	uint64_t charmap[4], size_t *count)
```
katef, line 99: really nice to have this in the api, thanks

katef (approved):

PR #493, opened 2024-09-12
bugfix: fsm_endid_get should sort with result count, not buffer size.

silentbicycle:
fsm_endid_get's id_buf_count argument is expected to "have enough cells (according to id_buf_count)", but if it has more than enough, stale data can get sorted into the result.

Add a test, tests/endids/endids_reused_buffer.c


katef (approved):

katef:
Thanks!

PR #496, opened 2024-09-18
fsm_compact_states must remap endids, to avoid dangling references.

silentbicycle:
fsm_compact_states must remap endids, to avoid dangling references.

Add a regression test showing possible endid false negatives when FSMs were trimmed (called from fsm_minimise) without updating endids.

### DIFF
--- a/src/libfsm/endids.c
+++ b/src/libfsm/endids.c
@@ -880,6 +880,8 @@ fsm_endid_iter_bulk(const struct fsm *fsm,
 		return 1;
 	}
 
+	const fsm_state_t state_count = fsm_countstates(fsm);
+
 	bucket_count = ei->bucket_count;
 
 	for (b_i = 0; b_i < bucket_count; b_i++) {
@@ -891,6 +893,7 @@ fsm_endid_iter_bulk(const struct fsm *fsm,
 
 		count = b->ids->count;
 
+		assert(b->state < state_count);
 		if (!cb(b->state, &b->ids->ids[0], count, opaque)) {
 			return 0;
 		}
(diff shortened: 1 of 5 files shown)


tests/endids/endids_trim_must_compact_endids.c
```diff
@@ -0,0 +1,64 @@
...
+
+	/* Create FSMs with a bunch of garbage states, to increase
+	 * the likelihood that the endid's state not being updated
+	 * after calling fsm_trim triggers the bug. */
+	for (size_t end_state = 1; end_state < limit; end_state++) {
+		struct fsm *fsm = fsm_new(NULL);
+		assert(fsm != NULL);
+
+		for (size_t i = 0; i < limit; i++) {
+			if (!fsm_addstate(fsm, NULL)) {
```
silentbicycle, line 25: `fsm_new_statealloc` is in `include/fsm/fsm.h` but not `libfsm.syms`.

src/libfsm/endids.c
```diff
@@ -880,6 +880,8 @@ fsm_endid_iter_bulk(const struct fsm *fsm,
 		return 1;
 	}
 
+	const fsm_state_t state_count = fsm_countstates(fsm);
```
katef, line 883: let's have this inside `#ifndef NDEBUG` if it's just here for the assertion

tests/endids/endids_trim_must_compact_endids.c
katef: We can expose it!

src/libfsm/endids.c
silentbicycle: Fixed in 8487fcb1.

tests/endids/endids_trim_must_compact_endids.c
silentbicycle: Added in 2d61cf94.

silentbicycle:
The fuzzer failed because it rediscovered the `(*:` case (#386).

PR #497, opened 2024-09-23
print API: Box end_ids and end_id_count in a struct for callbacks.

silentbicycle:
Move the end_id array and its count into a struct for state metadata, and rename access throughout to end_ids and end_id_count.

Upcoming changes for eager output IDs will soon be passing more info to all of these callbacks, but only callers making use of those fields need to care. Instead of making callers add more `(void) param;` declarations all over the place to avoid warnings, just pass in a metadata struct pointer. Also, "count" is a pretty generic name and what it refers to will soon be ambiguous.

This should not be a functional change on its own, it's just clearing the way for new features.

### DIFF
--- a/src/libfsm/print/ir.h
+++ b/src/libfsm/print/ir.h
@@ -54,8 +54,10 @@ struct ir_error {
 struct ir_state {
 	const char *example;
 
-	fsm_end_id_t *ids; /* NULL -> 0 */
-	size_t count:31; // :31 for packing
+	struct ir_state_endids {
+		fsm_end_id_t *ids; /* NULL -> 0 */
+		size_t count;
+	} endids;
 
 	unsigned int isend:1;
 
--- a/include/fsm/print.h
+++ b/include/fsm/print.h
@@ -42,6 +42,11 @@ enum fsm_print_lang {
 	FSM_PRINT_VMOPS_MAIN
 };
 
+struct fsm_state_metadata {
+	const fsm_end_id_t *end_ids;
+	size_t end_id_count;
+};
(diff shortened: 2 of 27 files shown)


src/libfsm/print/ir.h
```diff
@@ -54,8 +54,10 @@ struct ir_error {
 struct ir_state {
 	const char *example;
 
-	fsm_end_id_t *ids; /* NULL -> 0 */
-	size_t count:31; // :31 for packing
+	struct ir_state_endids {
+		fsm_end_id_t *ids; /* NULL -> 0 */
```
silentbicycle, line 58: This loses the `:31 for packing`, but if the goal is to save memory per IR state, then this could instead be a separately allocated struct like `struct ir_state_endids { size_t count; fsm_end_ids_t ids[]; }` and states without endids (which will be most states) save storing the count entirely. Thoughts?

silentbicycle:
I also moved it into a struct in the IR, because the same issues apply there -- it's about to get a different ID type (eager output IDs) and "ids" and "count" bare are ambiguous.

silentbicycle:
This failed due to the fuzzer hitting an unrelated bug, for `(?<m`. It may not be exactly the same bug as #386 but probably closely related if not.

src/libfsm/print/ir.h
katef: No need, the packing isn't that important.

katef (approved):

katef:
I might come back and rename this, the names are all too verbose for me...

PR #499, opened 2024-10-11
Randomized generation and misc. fixes

silentbicycle:
Instead of always using the first character in a label set, pseudorandomly pick one using `rand()` if a non-zero seed is provided. This adds an extra argument to `fsm_generate_matches`, for the seed.

Also:
- Remove a stale comment.
- Default `TRACK_TIMES` to 0 when `BUILD_FOR_FUZZER` is set, because it fills up the fuzzer logs with lots of uninteresting messages about things taking longer when subjected to weird random inputs.
- Only run `fsm_trim`'s integrity checks when `EXPENSIVE_CHECKS` is set.

### DIFF
--- a/src/libfsm/gen.c
+++ b/src/libfsm/gen.c
@@ -107,7 +107,7 @@ struct gen_ctx {
 static bool
 gen_init_outer(struct fsm *fsm, size_t max_length,
     fsm_generate_matches_cb *cb, void *opaque,
-    bool randomized, unsigned seed);
+    bool randomized);
 
 static bool
 gen_init(struct gen_ctx *ctx, struct fsm *fsm);
@@ -140,7 +140,7 @@ static bool
 grow_stack(struct gen_ctx *ctx);
 
 int
-fsm_generate_matches(struct fsm *fsm, size_t max_length, unsigned seed,
+fsm_generate_matches(struct fsm *fsm, size_t max_length, int randomized,
     fsm_generate_matches_cb *cb, void *opaque)
 {
 	if (max_length == 0) {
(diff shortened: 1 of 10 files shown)


src/libfsm/gen.c
```diff
@@ -208,8 +213,9 @@ gen_init_outer(struct fsm *fsm, size_t max_length,
 
 	assert(fsm_all(fsm, fsm_isdfa)); /* DFA-only */
 
-	assert(!randomized);		 /* not yet supported */
-	(void)seed;
+	if (randomized) {
+		srand(seed);
+	}
```
katef, line 218: Not sure about calling `srand()` in library code, because of its state being side-effect to the caller's own uses. I don't want to include our own prng. What do you think of moving the `srand()` invocation to the caller?

silentbicycle: Good point. Should the interface just have `bool randomized` (or `int`, since nothing else in that header uses `bool`), note that it uses `rand` specifically, and encourage the caller to seed with `srand`? It doesn't need to be a _good_ PRNG, just something that has more variation than always choosing the first printable character.

silentbicycle: Updated in cf8fc651.

katef (approved):

src/re/main.c
```diff
@@ -1047,7 +1047,7 @@ main(int argc, char *argv[])
 	}
 
 	if (generate_bounds > 0) {
-		if (!fsm_generate_matches(fsm, generate_bounds, fsm_generate_cb_printf_escaped, &opt)) {
+		if (!fsm_generate_matches(fsm, generate_bounds, 0, fsm_generate_cb_printf_escaped, &opt)) {
```
katef, line 1050: would make sense to expose this to the cli of course (and also for fsm(1)), but that can wait

PR #500, opened 2024-10-12
Add re_is_anchored interface.

silentbicycle:
Putting this in its own PR, as requested.

### DIFF
--- a/include/re/re.h
+++ b/include/re/re.h
@@ -136,6 +136,22 @@ re_comp(enum re_dialect dialect,
 	const struct fsm_alloc *alloc,
 	enum re_flags flags, struct re_err *err);
 
+/* Parse and analyze the regex enough to determine whether it is
+ * anchored at the start and/or end.
+ *
+ * As long as the result is checked for RE_IS_ANCHORED_ERROR first,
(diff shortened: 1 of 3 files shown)


include/re/re.h
```diff
@@ -136,6 +136,22 @@ re_comp(enum re_dialect dialect,
...
+	RE_IS_ANCHORED_NONE = 0x00,
+	RE_IS_ANCHORED_START = 0x01,
+	RE_IS_ANCHORED_END = 0x02,
+	RE_IS_ANCHORED_BOTH = 0x03,
+	RE_IS_ANCHORED_ERROR = 0xFFFF,
+};
+enum re_is_anchored_res
+re_is_anchored(enum re_dialect dialect, re_getchar_fun *f, void *opaque,
+	enum re_flags flags, struct re_err *err);
+
```
katef, line 154: this ought to be exposed to re(1) as a CLI interface in a similar way to `re_is_literal`. Which means tests for the CLI tool, and manpage stuff too.

I'm going to :+1: this PR now, we can come back and add those in the future.

katef (approved):

PR #504, opened 2024-11-08
Fix a memory leak during `fsm_determinise_with_config`'s early exit

silentbicycle:
- Add a `tests/regressions` directory for test cases that need a standalone C file, because they aren't easily reproducible via fsm, re, etc. input files.
- Add a test case that exercises `fsm_determinise_with_config`'s early exit due to the state limit. When run with LeakSanitizer or Valgrind and with the previous commit reverted, it will show that the edge sets for the DFA being constructed weren't freed properly.
- Explicitly set m->edges to NULL when transferring ownership from the stack/map to the DFA, and free any edge sets in the cleanup handler that are still in the map.

### DIFF
--- a/tests/regressions/regressions_determinise_state_limit_leak.c
+++ b/tests/regressions/regressions_determinise_state_limit_leak.c
@@ -1,0 +1,69 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+
+#include <assert.h>
+
+#include <re/re.h>
+#include <fsm/fsm.h>
+#include <fsm/bool.h>
+
+static const char *strings[] = {
+	[0] = "apple",
+	[1] = "banana",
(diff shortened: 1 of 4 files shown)


tests/regressions/regressions_determinise_state_limit_leak.c
```diff
@@ -0,0 +1,69 @@
...
+#include <stdbool.h>
+#include <string.h>
+
+#include <assert.h>
+
+#include <re/re.h>
+#include <fsm/fsm.h>
+#include <fsm/bool.h>
+
+static const char *strings[] = {
```
silentbicycle, line 12: The input here is arbitrary, but one easy way to make a FSM whose state count increases significantly during determinisation is to combine several unanchored patterns.

katef (approved):
Thanks!

PR #505, opened 2025-02-11
Update CI to use Ubuntu 22.04 

deg4uss3r:
CI has been failing since GitHub updated the default Ubuntu runners to 24.04 which gets a very new version of Clang (18+), this locks CI to Ubuntu 22.04 and clang at 14.0.0 so we can still verify builds work.


katef (approved):
Thank you!

PR #509, opened 2025-07-03
Fix lx token identification

silentbicycle:
Changing the leaf and endleaf callbacks to accept and reject in #485 broke lx, but it went unnoticed for a while. This fixes it.

libfsm's normal execution mode evaluates a DFA, character by character, terminating either when the next character isn't a valid edge or the end of input is reached (in which case it checks end state metadata). lx's execution mode is a little different, because it's tokenizing -- instead of reading to the end of input, it should consume as much consecutive input that matches a particular token, then push back the last character read (so it can resume with it as context for the next token), yield the token type, and suspend.

lx used to work by breaking abstraction and calling directly into `fsm_print_cfrag` (overriding the leaf behavior to yield token types, and adding an extra 'NONE' state to the generated state machine code), but when the callback interfaces shifted its internals no longer fit what lx expected. Now the reject hook is passed the same state metadata as the accept state, and the reject hook in lx checks whether the end id is associated with a particular AST mapping and token type.

This is only implemented for the "c" output format, but similar changes could possibly make others usable without a lot more work. In particular, kate mentioned it'd be good to be able to use "vmc" output instead of "c" moving forward.

Most of the code changes happen inside of lx's code generation, but there are a few elsewhere:

- The reject hook now has a state_metadata pointer, so update the callers for all the output formats.

- libfsm's 'c' output now includes a macro `FSM_ADVANCE_HOOK(C)`, which is called with the next character read in the FSM_IO_STR and FSM_IO_PAIR io modes immediately after advancing. This is used to inform lx's internal bookkeeping about token positions and buffering token names. FSM_IO_GETC doesn't need it, because its getc callback manages the character stream. The macro defaults to a no-op when undefined.

- libfsm's 'c' output also includes a flag, `has_consumed_input`, so the code expanded in place from the reject/accept hooks can determine when the state machine input handler loop has consumed any input. This was previously encoded by the extra NONE state.

lx's code generation using this flag is a bit cluttered, because the reject hook doesn't know whether it's expanding for the end states, but it's probably not worth changing the reject hook type signature to add another flag. This results in checks for has_consumed_input in code paths where trivial static analysis would show it to be dead code, and some extra unreachable code at the end of the function.

Also:

- Re-enable the lxpos tests (using `getc`, `str`, and `pair`), and fix a couple things in their build config.
- Make lx's `-l dump` output call `lx->free` when necessary.
- Make sure the makefile test targets check `*res*` not `res*`, because some result files are prefixed.

There may still be combinations of lx output modes and other flags that get warnings for unused functions or otherwise have a broken build, but I tested `getc`, `str`, and `pair`, with `dyn` or `fixed` buffering, and with or without `-x buf -x pos`. All the combinations tested by the lxpos build work.


src/libfsm/print/c.c
```diff
@@ -478,6 +484,41 @@ fsm_print_c_body(FILE *f, const struct ir *ir,
...
+	 * character logic expanded here.
+	 *
+	 * lx needs to track this for proper EOF handling. It previously
+	 * generated the state enum itself, so that it could include an
+	 * additional 'NONE' state. Inside the input loop, the default
+	 * state of NONE would be updated to the start state, but if the
+	 * input loop was skipped it would still be NONE. */
+	fprintf(f, "\tint has_consumed_input = 0;\n");
+
+	/* For FSM_IO_STR and FSM_IO_PAIR, define a macro that will be
```
silentbicycle, line 498: This macro is a bit ugly, but much simpler than changing normal code generation to instantiate some kind of character iterator, just so lx and libfsm are on the same page about advancing and pushing back the character stream.

src/lx/print/c.c
```diff
@@ -371,54 +501,39 @@ print_io(FILE *f, const struct fsm_options *opt)
...
-	case FSM_IO_STR:
-		/*
-		 * For FSM_IO_STR we treat '\0' as the end of input,
-		 * and so there's no need to distinguish it from EOF.
-		 * We return '\0' here to save the assignment.
-		 */
-		fprintf(f, "\tassert(lx->p != NULL);\n");
+		/* FIXME: This should distinguish between alloc failure
+		 * and EOF, but will require layers of interface changes. */
+		fprintf(f, "\tif (!lx_advance_end(lx, c)) { return EOF; }\n");
```
silentbicycle, line 507: `lx_advance_end` can fail if `lx->push` has an allocation failure, but `lx_getc` either returns a character or EOF. It should probably signal the error more obviously, but what do you think the interface should look like?

src/lx/print/c.c
```diff
@@ -599,49 +743,33 @@ print_buf(FILE *f)
...
 	assert(z != NULL);
 	assert(z->fsm != NULL);
 	assert(fsm_all(z->fsm, fsm_isdfa));
 	assert(ast != NULL);
-	assert(cp != NULL);
+	assert(cur_char_var != NULL);
 
-	/* TODO: prerequisite that the FSM is a DFA */
+	/* prerequisite that the FSM is a DFA */
+	assert(fsm_all(z->fsm, fsm_isdfa));
```
silentbicycle, line 758: implemented this

silentbicycle:
I added some negative tests for truncated/corrupt input getting error messages and found that the block comment in `in3.txt` isn't reported as an error if the `*/` is missing. I will add the tests to the PR once I have that fixed.

silentbicycle:
I called it out in the commit logs for ae53e94, but this changes the result for `test/lxpos/in6.txt`. Now it explicitly fails tokenization when there's an EOF before matching the second pattern in e.g. `'//` .. /\n/ -> $nl`, rather than just failing to yield `TOK_NL` but still yielding `TOK_EOF`.

src/lx/print/c.c
```diff
@@ -137,13 +136,37 @@ shortest_example(const struct fsm *fsm, const struct ast_token *token,
...
+	}
+}
+
+static void
+unget_character(FILE *f, bool pop, const char *cur_char_var)
+{
+	fprintf(f, "%sungetc(lx, %s); ", prefix.api, cur_char_var);
+	if (pop && (~api_exclude & API_POS)) {
+		fprintf(f, "%s%spop(lx->buf_opaque); ",
+		    prefix.api, buf_op_prefix());
```
katef, line 158: sensible, okay

src/lx/print/c.c
```diff
@@ -152,28 +175,51 @@ accept_c(FILE *f, const struct fsm_options *opt,
...
-		fprintf(f, "lx->z(lx)");
+		if (m->token == NULL) {
+			/* update to a different zone, then call to it */
+			fprintf(f, "lx->z = z%u, lx->z(lx)", zindexof(ast, m->to));
+		} else {
+			/* update zone, then yield a token */
+			fprintf(f, "lx->z = z%u, ", zindexof(ast, m->to));
+			fprintf(f, "%s", prefix.tok);
+			esctok(f, m->token->s);
+		}
```
katef, line 214: okay, cool

src/libfsm/print/c.c
```diff
@@ -222,10 +222,16 @@ print_case(FILE *f, const struct ir *ir,
 	assert(f != NULL);
 	assert(cs != NULL);
 
+	assert(state_id < ir->n);
+	const struct fsm_state_metadata state_metadata = {
+		.end_ids = ir->states[state_id].endids.ids,
+		.end_id_count = ir->states[state_id].endids.count,
+	};
+
```
katef, line 230: why not pass whatever type `&ir->states[state_id]..endids` is?

src/libfsm/print/c.c
katef: This macro rubs me the wrong way too. There's gotta be a nicer way. Maybe we could have lx emit a call to `%sadvance_end()` somewhere (in a new hook, e.g. at the end of the generated fsm loop) instead.

src/libfsm/print/c.c
silentbicycle: `struct ir_state_endids` is defined in an internal `ir.h` header, and an upcoming PR is going to add another pair of fields to the state metadata struct, for eager output IDs and their count.

src/lx/print/c.c
```diff
@@ -182,7 +228,58 @@ reject_c(FILE *f, const struct fsm_options *opt,
...
+		} else if (m->token == NULL && m->to == NULL && has_endids) {
+			unget_character(f, true, env->cur_char_var);
+		} else if (m->token == NULL && m->to != NULL) {
+		  	unget_character(f, true, env->cur_char_var);
+		} else if (m->token != NULL && m->to == NULL) {
+			assert(has_endids);
+		  	unget_character(f, true, env->cur_char_var);
+		} else if (m->token != NULL && m->to != NULL) {
+		  	unget_character(f, true, env->cur_char_var);
+		}
```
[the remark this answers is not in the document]
silentbicycle: I was making the cases explicit. They're almost all handled the same way though, so I could collapse that logic a bit. Does that seem worth doing? This module should get replaced with the vmc codegen soon, though.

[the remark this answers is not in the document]
katef: Seems fine

src/libfsm/print/c.c
```diff
@@ -478,6 +484,41 @@ fsm_print_c_body(FILE *f, const struct ir *ir,
...
+	/* This flag indicates whether the any of the input stream was
+	 * consumed before getting EOF and skipping over the state and
+	 * character logic expanded here.
+	 *
+	 * lx needs to track this for proper EOF handling. It previously
+	 * generated the state enum itself, so that it could include an
+	 * additional 'NONE' state. Inside the input loop, the default
+	 * state of NONE would be updated to the start state, but if the
+	 * input loop was skipped it would still be NONE. */
+	fprintf(f, "\tint has_consumed_input = 0;\n");
```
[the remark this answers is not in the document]
katef: this makes me so happy, thank you!

src/lx/print/c.c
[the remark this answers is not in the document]
katef: uh i should translate "fine" from british to american: i mean i like it

PR #520, opened 2026-07-09
Flags for ${...} syntax for `re_interpolate`

katef:
Exciting things I do here:
- rename to `re_interpolate()`, I decided I just like the shorter name. I'm keeping it in <re/groups.h> because I expect other groups-related things to land there
- `RE_INTERPOLATE_SINGLE_DIGIT` treats subsequent digits after the first as literals. so e.g. `\123` means `\1` followed by a literal "23"
- `RE_INTERPOLATE_BRACES` allows for writing `${11}1` meaning group 11 followed by a literal "1"

named captures would also go inside the `{...}` syntax, but I don't want to introduce that yet.


june-fastly (approved):
looks good

cxreg (approved):
This seems understandable and correct to me, thanks!

